PR #24, opened 2015-08-12
add hiring button to the end of blog posts

jukecraft:
I'm not happy with it being positioned so far on the right, but I remember what a pain it was to make the button on the front page work across devices and browsers, so I don't really want to touch the css.

### DIFF
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -38,7 +38,7 @@ group:
         <hr/>
         <p class="cross-post"><a href="{{ page.canonical.href }}">This post was cross-posted to {{ page.canonical.name }}.</a></p>
         {% endif %}
-      <a class="btn-hire btn-u-orange ls-s1" href="/careers/craftsman/">We are hiring!</a>
+        <p><a class="btn-hire btn-u-orange ls-s1" href="/careers/craftsman/">We are hiring!</a></p>
     </div>
 
     <!--End Blog Post-->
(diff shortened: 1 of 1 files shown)


franziskas:
I'm not happy with it being positioned so far on the right, but I remember what a pain it was to make the button on the front page work across devices and browsers, so I don't really want to touch the css.

_layouts/post.html
```diff
@@ -38,7 +38,9 @@
         <hr/>
         <p class="cross-post"><a href="{{ page.canonical.href }}">This post was cross-posted to {{ page.canonical.name }}.</a></p>
         {% endif %}
+      <a class="btn-hire btn-u-orange ls-s1" href="/careers/craftsman/">We are hiring!</a>
```
SamirTalwar: Links are inline elements. Should it be in a `<p>`, `<div>` or similar so we don't have a mix of block and inline elements in the containing `<div>`?

SamirTalwar:
Blog hiring button

This looks… massive. Can we shrink it down? And maybe lose the dead space on the right of the text?

Maybe it's not supposed to look like that. I tried it using the latest versions of Firefox and Chrome on Mac.

jukecraft:
As I said in the first comment, I don't really want to mess with _this_ version of the button, as it took a while to get it working for the front page on different devices/browsers. This is a quick fix until someone is happy to make another button that fits better :)

PR #41, opened 2015-08-26
Better links in bullet point lists

jukecraft:
make links show up as defined in orange.css even when they are part of bullet points

### DIFF
--- a/assets/css/style.css
+++ b/assets/css/style.css
@@ -37,26 +37,25 @@
 /*Import Google Font*/
 @import url(http://fonts.googleapis.com/css?family=Open+Sans:300,400);
 
-body { 
-   color: #333; 
+body {
+   color: #333;
    font-size: 13px;
    line-height: 1.6;
 }
 
-p, 
-li,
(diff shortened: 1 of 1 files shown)


robertfirek:
I'm not sure if **_assets/css/style.css**_ is the part of Unify Bootstrap template. Modifying this file will not allow us to upgrade to the latest version of the template in the future.

jukecraft:
It's a quick fix Mash came up with. We don't have a long term solution for these kinds of problems, if you have an idea, go ahead :)

robertfirek:
We should have our own custom css file at the end of list of all css stylesheets. This stylesheet can override styles defined in the previous files.

jukecraft:
We have a custom.css but we have not decided on how to deal with this. You can suggest something.

robertfirek:
I think we can move all changes from style.css to custom.css. It should
change style on the site in the same way.

jukecraft:
The problem is that I didn't add something, I removed something. How would we do that?

PR #43, opened 2015-09-01
Change heading level for headline

tekavec:
Unify template has a SEO design flaw - it uses h2 heading for the top headline for all inner pages at the breadcrumbs. Each page should have one h1 headline, according to SEO best practices, so I changed h2 to h1. Further headings hierarchy of particular inner pages should be examined (h1 should be followed by h2, h3 etc.).

### DIFF
--- a/_includes/breadcrumb.html
+++ b/_includes/breadcrumb.html
@@ -1,7 +1,7 @@
 {% if page.title or page.tag %}
     <div class="breadcrumbs margin-bottom-10">
         <div class="container">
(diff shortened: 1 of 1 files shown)


tekavec:
Some further reading:
- http://www.w3.org/MarkUp/html-spec/html-spec_5.html#SEC5.4
- http://www.hobo-web.co.uk/headers/
- http://www.searchenginejournal.com/in-2014-how-important-is-an-h1-tag-for-seo/

PR #113, opened 2016-01-18
Spring integration tests

dhatanian:


_posts/2016-01-18-isolating-integration-tests-with-the-spring-framework.md
```diff
@@ -0,0 +1,141 @@
...
+ * `@WebIntegrationTest`: Tells Spring that this is an integration test running a web application, otherwise by default Spring will not run an HTTP server in test mode. We also set the `server.port` to a value of `0` so that Spring Boot choose a random port for the HTTP server to listen to. This allows to have several tests running in parallel, or to have another version of the application running in the background.
+ * `@SpringApplicationConfiguration`: We tell Spring where it will find the `DelayedWebServerConfiguration` class we created before.
+ * `@ActiveProfiles`: Enables the `restTemplateTimeout` profile, otherwise the `Controller` and the `Configuration` will be filtered out.
+
+We now have a unit test running with a limited set of dependencies instead of the whole application. What if we wanted to go further and add mocks into the game? This may be required when a dependency does not have a dev environment or that it is too complicated to call from a developer's workstation. In that case, we can add those mocks to the `Configuration` class and they will be injected into the test's Spring context.
```
mashooq: should be "integration test" not "unit test"

_posts/2016-01-18-isolating-integration-tests-with-the-spring-framework.md
```diff
@@ -0,0 +1,141 @@
...
+
+  @Autowired
+  private RestTemplate restTemplate;
+
+  @Test
+  public void throw_timeout_if_response_lasts_more_than_two_seconds() {
+    thrown.expect(ResourceAccessException.class);
+    thrown.expectCause(instanceOf(SocketTimeoutException.class));
+
+    callEndpointWithDelay(3000);
```
mashooq: This test may hang indefinitely in the failure scenario. I suggest that you wrap the call in using awaitility to time it out on your end if it doesn't on the server. (e.g. await().atMost(4000, MILLISECONDS).until(callEndpointWithDelay(3000)); )

_posts/2016-01-18-isolating-integration-tests-with-the-spring-framework.md
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  @Test
+  public void do_not_throw_timeout_if_response_lasts_less_than_two_seconds() {
+    callEndpointWithDelay(10);
```
mashooq: This may also hang in failure scenario like the other one.

_posts/2016-01-18-isolating-integration-tests-with-the-spring-framework.md
roberttaylor426: I don't think you want to use Awaitility in this case (it would repeatedly call `callEndpointWithDelay(3000)` every 100 milliseconds by default). Isn't the standard [JUnit timeout](https://github.com/junit-team/junit/wiki/Timeout-for-tests) mechanism what you're after?

mashooq: Good point. I had neglected the fact that it polls.

dhatanian: @mashooq When you talk about a failure scenario, do you mean the failure of the `DelayedWebServerController` ?

If the timeout is not correctly set, it will not hang indefinitely since the controller returns after a few seconds.

_posts/2016-01-18-isolating-integration-tests-with-the-spring-framework.md
dhatanian: Again, unless you foresee a failure of the `DelayedWebServerController` I don't see how this could fail indefinitely since the controller returns?

PR #114, opened 2016-01-19
ShareThis Fix

TobyRet:
Tweaks to shareThis buttons so that the link that shared is the actual blog post, not the home page.


jukecraft:
share button seems to still not point to the correct URL on pedro’s new blog

(the twitter one) https://shar.es/1hgPV8

PR #120, opened 2016-01-29
Add correct link to Sandro's book on Amazon

TobyRet:


jukecraft:
YESSSS Finally thank you!

PR #134, opened 2016-02-23
Suggestion for new blog title

TobyRet:
This is a suggestion only, in order to make the blog more 'clickable'.

The replacement 'Open Book Management' was influenced  by this wiki article here:
https://en.wikipedia.org/wiki/Open-book_management


sandromancuso:
I'll keep the title as is but I added a tag and a link to open book management. Thx.

PR #148, opened 2016-03-16
Removed header Introduction, added line-break and the 3 diagrams in the summary section

neomatrix369:
- Removed header Introduction as it looks strange in the snapshot blog page, added a line-break otherwise intro text appears too close to the tags
- Also added the three diagrams in the summary section - comparing each other, improving the visual experience


_posts/2016-03-16-Containers-all-the-way-through.md
```diff
@@ -24,7 +24,7 @@ tags:
 - operating system
 --- 
 
-### Introduction
+<br/>
```
robertfirek, line 27: Do you really need `<br>`? :)

neomatrix369, line 27: Yes otherwise the tags and intro text look squashed together, maybe changing something in the template could help

PR #165, opened 2016-05-04
Fixes to Async systems with sync clients.

felipefzdz:


_posts/2016-04-28-async-systems-with-sync-clients.md
```diff
@@ -12,12 +12,21 @@ image:
...
 
 ---
 
-As the [Reactive Manifesto](http://www.reactivemanifesto.org/) says Reactive systems are: Responsive, Resilient, Elastic and Message Driven. The last principle often goes together with non-blocking async protocols. This style of communication "allows recipients to only consume resources while active, leading to less system overhead". This fits perfectly with new demands of efficiency derived from the elastic model of cloud providers. However not every system is async and integrating an async system with a sync client could be tricky.
+As the [Reactive Manifesto](http://www.reactivemanifesto.org/) says Reactive systems are:
+
+* Responsive
+* Resilient
+* Elastic
+* Message Driven.
```
neomatrix369: remove the dot at the end

neomatrix369:
Just for curiosity sake its worth running this post via word cloud generator to see the terms used terms which in turn would help find out which categories the post belongs to most, try this http://www.tripwiremagazine.com/2012/12/12-cool-word-cloud-generators.html.

neomatrix369:
At the bottom of this post you can link it to the next post (Part 2) so people can read on.

felipefzdz:
Thanks for the tip, I didn't know about those tools. I've run it against the posts an current tags are ok, but it's pretty useful for next ones, thanks! I've applied the other two comments.

neomatrix369:
Sure, useful for thinking about higher level abstractions - its best not to use the results of cloud generators directly but to use them as an indicator of synonyms and use the words against a thesaurus. I'm glad it helped.

neomatrix369:
Looks good - i'll merge it.

PR #172, opened 2016-05-06
Language feature envy.

pedromsantos:


dhatanian:
@pedromsantos @stevelydford any reason why this branch is not deleted? 

Too used to managing source code in a shared network folder? :-D

PR #187, opened 2016-05-16
Move blog panel up, add sponsoring panel

Gryff:


Gryff:
Abandoning for now, as the new images look bad when the viewport is mobile-sized.

PR #224, opened 2016-06-12
Suggestion: Add a synonym for the term `mock`

alvarogarcia7:
add a synonym to mock: test double
reach a bigger audience, depending on the terms that they use


alvarogarcia7:
@sandromancuso could you please take a look?

services/training/crafting-code.html
```diff
@@ -85,7 +85,7 @@ <h4 class="margin-bottom-20">Created by Sandro Mancuso</h4>
                 <li>Are familiar with at least one Object-Oriented language</li>
                 <li>Are able to understand Java or C#</li>
                 <li>Can write and execute unit tests using a framework (such as JUnit, NUnit, etc.)</li>
-                <li>Have a basic understanding of mock objects</li>
+                <li>Have a basic understanding of mock objects or test doubles</li>
```
mashooq, line 88: Mock objects are a form of test doubles so if we have "test doubles" then we don't need to include "mock objects"

alvarogarcia7:
This PR has not gathered enough traction to be merged

PR #233, opened 2016-06-23
New blog - Single Responsibility Principle

TobyRet:


_posts/2016-06-23-Single-Responsibility-Principle.md
```diff
@@ -0,0 +1,108 @@
...
+- single responsibility
+- software design  
+---
+
+[SOLID](http://bit.ly/1Da1b16) (single responsibility, open-closed, Liskov substitution, interface segregation and dependency inversion) is a set of principles popularised by [‘Uncle’ Bob Martin](https://en.wikipedia.org/wiki/Robert_Cecil_Martin) that aim to guide developers in implementing good software design. The principles define how Object Oriented classes should relate to each other in order to create a codebase that is flexible and able to accommodate ongoing change.
+
+The first of these, the **Single Responsibility Principle** (SRP) states that -
+
+<blockquote class="hero">
+	<p><em>"Every module or class should have one, and only one reason to change."</em></p>
```
neomatrix369, line 23: There is this view of SRP as well "The single responsibility principle states that every module or class should have responsibility over a single part of the functionality provided by the software, and that responsibility should be entirely encapsulated by the class. All its services should be narrowly aligned with that responsibility." 

Either at this point or later on you want to save SRP is seen (or can be) from outside and inside. Maybe talk about both the views.

TobyRet, line 23: @neomatrix369 - Thanks for the comment. What is the source of the view you presented above?

neomatrix369, line 23: This was wikipedia and so this definition appears to me to be the outside view of SRP.

What you wrote appears to be the inside view.

I'll review further and give you more feedback, as SRP casts its net quite wide into all the different principles.

neomatrix369:
About the image, unless you are trying to be ironic about SRP I would use another one, the swiss knife is sort of an anti-pattern but its discussable.

_posts/2016-06-23-Single-Responsibility-Principle.md
TobyRet, line 23: Thanks for your comments @neomatrix369. One of the key messages of this article is that SRP can be prone to misinterpretation for a number of reasons however, definitions are not the main focus. I wanted to concentrate more on the practicalities of implementing SRP. Unless there are similar comments then I am inclined to think that discussing outside vs inside SRP might be out of scope for this article? I will update the source of my SRP quote - it's Uncle Bob, not wiki.

neomatrix369, line 23: Ok. As you are driving the post, you can choose the directions. I was just making you aware of the aspects.

PR #239, opened 2016-06-29
Redesign our practices

Gryff:
image

image


alvarogarcia7:
@Gryff when did you take the screenshots? how come the address is still the old one?

PR #273, opened 2016-07-14
Add 'Abstract section' in a post

alvarogarcia7:
Please wait for merging, until @TobyRet has approved the suggested abstract

### DIFF
--- a/_includes/related_posts.html
+++ b/_includes/related_posts.html
@@ -3,10 +3,10 @@
 <div class="container">
 
 		{% assign posts = site.posts %}
-        {% comment %}
-        Initializing arrays
-        You cannot initialize arrays using only Liquid. You can, however, use the split filter to break a string into an array of substrings.
-        {% endcomment %}
+		{% comment %}
+		Initializing arrays
+		You cannot initialize arrays using only Liquid. You can, however, use the split filter to break a string into an array of substrings.
+		{% endcomment %}
 		{% assign relatedPosts = ""  %}
 			  
 		{% for post in posts %}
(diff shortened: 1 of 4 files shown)


alvarogarcia7:
@TobyRet has now approved this comment

_includes/related_posts.html
```diff
@@ -51,6 +58,13 @@
 					{% assign author = relevantPostAttributes[5] %}
 					{% assign postContent = relevantPostAttributes[6] %}
 					{% assign postDate = relevantPostAttributes[7] %}
+					{% assign abstract = relevantPostAttributes[8] %}
+					
+					{% if abstract %}
```
Gryff, line 63: @alvarogarcia7 Just to double check, relevantPostAttributes[8] doesn't error if there is nothing there? So all other posts without abstracts aren't affected.

alvarogarcia7, line 63: @Gryff good eye. it does not fail, but `abstract` is null/empty and does not pass the check on line 63: `if abstract`.

Gryff, line 63: Awesome, thanks :)

Gryff, line 63: @alvarogarcia7 One last thing is to check with @halimakoundi as there may be some merge conflicts here?

ghost, line 63: @alvarogarcia7 and @Gryff, the change I made in that file is at line 5, so git should be able to merge those two without issues.. right?

Gryff, line 63: Sounds correct to me :)

PR #274, opened 2016-07-15
Configure 404 page for analytics

alvarogarcia7:
https://trello.com/c/uDfRQF4i/8-adding-analytics-to-the-404-page-so-that-we-track-dead-links

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -53,6 +53,8 @@
 
     ga('create', '{{site.googleanalytics.account}}', 'codurance.com');
     ga('send', 'pageview');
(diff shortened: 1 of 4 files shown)


_includes/head.html
```diff
@@ -53,11 +53,13 @@
 
     ga('create', '{{site.googleanalytics.account}}', 'codurance.com');
     ga('send', 'pageview');
+    window.googleAnalyticsEvents = { environment: {ga:ga}, functions: []};
```
Gryff, line 56: @alvarogarcia7 Should all GA reporting move to this events model in future? (I don't have any experience implementing GA on a website, no idea what best practices are)

PR #286, opened 2016-07-28
Add The Software Craftsman to Our Company page

Gryff:
The page itself has not been redesigned so I didn't attempt to change the design of the page, only add Sandro's book in similar design. 

image

### DIFF
--- a/aboutus/ourcompany.html
+++ b/aboutus/ourcompany.html
@@ -35,36 +35,47 @@ title: Our Company
             </div>
 
             <div class="row">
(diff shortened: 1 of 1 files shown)


alvarogarcia7:
@Gryff I'm missing a link to the SC manifesto, maybe at the end of the first paragraph

Gryff:
@alvarogarcia7 Good point. It's unclear from the PR because I did some formatting, but I didn't write this copy, it is unchanged from the old version. I just added the book on the right. 
I'll add a link to the manifesto now.

PR #296, opened 2016-08-08
Blog cta design

ghost:
blog page - call to action


assets/css/custom.css
```diff
@@ -424,7 +424,7 @@ body:last-child .btn-u, x:-moz-any-link {
 .browse-courses {
```
tekavec, line 418: Could we reuse this style in other places as well? If this is the case, a more generic name would be more appropriate than content related.

assets/css/custom.css
```diff
@@ -424,7 +424,7 @@ body:last-child .btn-u, x:-moz-any-link {
...
     display: inline-block;
     border-radius: 6px;
```
tekavec: Border radius is used in 8 different places. We can do it more DRY if we define a single style (e.g. ".rounded") and add it to all elements that require it. If we want to change the rounding value in the future, we'll do it in one single place.

assets/css/custom.css
```diff
@@ -424,7 +424,7 @@ body:last-child .btn-u, x:-moz-any-link {
...
-    margin: 20px 0px 0px 10px;
+    margin: 20px 10px 12px;
```
tekavec: It's worth trying implement vertical alignment with table and table-cell display values. It's more stable, e.g. changing the font size shouldn't break the vertical alignment. Example:
html:

```
<div class="parent">
  <div class="child">Content here</div>
</div>
```

css:
.parent {display: table}

.child {
  display: table-cell;
  vertical-align: middle;
 }

ghost: I tried to use table and table cell display attributes but the problem is that it will break other elements on other pages where the class contact-us is used.
I would need to call the elements on the blog page differently and it seems wrong (have them use a different class)

tekavec: Good concern! I lack the context so I'll make some guess:
- for the sake of consistency (and if all "contact-us" containers have similar purpose) the same styling for the text and link buton could be used on the blog page; moreover, there are indeed two strong sections below blog pagination section, which could be more visually separated: comment/question form and link to courses
- if it really has to be visually different, then it deserves of course a different container style (but keep in mind the inner context will drop in two lines at least on mobile view)

PR #297, opened 2016-08-11
Upgrade jekyll

mashooq:
Requires Ruby 2.3.1 ('''rvm install ruby-2.3.1''')
- then '''bundle install'''
- then '''rake serve'''


felipefzdz:
I've reviewed the web after the upgrade and I couldn't find anything wrong, so I'll merge the PR.

PR #303, opened 2016-08-16
added a software craftsmanship section and removed the random our values bit

mashooq:
home-page-capture


index.html
```diff
@@ -209,54 +209,40 @@ <h2 class="gray-underline">How We Work</h2>
...
-            <div class="col-md-4 md-margin-bottom-30">
-                <i class="fa fa-trophy fa-3x"></i>
-                <div class="content-box">
-                    <h3 class="heading-sm">QUALITY</h3>
-                    <p>We don't negotiate on quality. Software must deliver more value with time and not be a source of increasing pain and cost.</p>
+            <div class="row news-v1">
+                <div class="col-md-4 md-margin-bottom-40">
+                    <div class="news-v1-in">
+                        <img class="img-responsive" src="/assets/img/custom/home/the-software-craftsman-book.jpg" alt="">
+                        <h3><a href="#">The Software Craftsman: by our Co-founder Sandro Mancuso</a></h3>
```
CarlosMChica: Do we really want to scroll to the top of the page when the link is clicked? 
I think we have two options:
1. Remove the link here and use the one at the bottom.
2. Include the link here and remove the link at the bottom

The same thing applies to the other columns' headers anchors

index.html
```diff
@@ -209,54 +209,40 @@ <h2 class="gray-underline">How We Work</h2>
...
+                        <p>The book is described as "an encyclopedia on the behaviours, attributes, and structure of an organisation that adheres to principles of Software Craftsmanship". A must read for anyone wanting to better understand the values behind the movement.</p>
+                        <ul class="list-inline news-v1-info no-margin-bottom">
+                            <li><a href="https://www.goodreads.com/book/show/23215733-the-software-craftsman">Discover more on Good Reads ...</a></li>
```
CarlosMChica, line 222: Is there a reason because we don't open the link in a new tab? 

I just checked some of the links in the site's footer and all of them open in a new tab. However, the previous "We are proud organisers of" links open in the same tab.

mashooq, line 222: Generally - opening links in a new tab is not good user experience. Although I don't feel too strongly about this in this case because it's a separate website. Lets stick with them as is for now and change later if others have a stronger opinion.

index.html
mashooq: Added the same link to the headings now.

PR #320, opened 2016-08-25
Use abstract when exist to display latest blog

ghost:


index.html
```diff
@@ -27,7 +32,7 @@
                         {% endif %}
                         <div class="blog-content">
                             <h3><a href="{{ post.url }}">{{ post.title | truncatewords: 8 }}</a></h3>
-                            <p>{{ post.content | strip_html | truncatewords:30 }}</p>
+                            <p>{{ introductionToPost | strip_html | truncatewords:30 }}</p>
```
CarlosMChica, line 36: For consistency sake, I'd say that following what is done in the lines 28-32 with the `img` tag, the if to decide what introduction to used could be inlined here. Another option could be to separate the if statement for the `img` tag.

ghost: this is now aligned
thannks

PR #322, opened 2016-08-25
Flatten services page

CarlosMChica:


CarlosMChica:
Don't merge yet. Need to remove old services section.

PR #338, opened 2016-09-01
Fix old reference to missing resource and remove unused css

CarlosMChica:


CarlosMChica:
- There were plenty of different CSS files for header and footer. Got rid of the unused ones

PR #344, opened 2016-09-02
Add fix to prevent firing navigation on team member link click

CarlosMChica:


CarlosMChica:
This is the only way I've been able to block the click event to navigate. If there is a better way please let me know :)

PR #363, opened 2016-09-14
Add Spanish translation for services page

CarlosMChica:


_i18n/es.yml
```diff
@@ -140,6 +140,47 @@ publications-tweet_feed:
 publications-post:
    author_title: Por
    date_title: Publicado
+services-software_creation:
+   title: Creación de software
+   subtitle: <p><strong>Te ayudaremos a crear software bien elaborado que aporta valor a su negocio.</strong></p>
```
alvarogarcia7: Difference in style: tú vs usted. please pick one and keep it consistent

_i18n/es.yml
```diff
@@ -140,6 +140,47 @@ publications-tweet_feed:
...
+           <p>Nuestros craftsmen han perfeccionado su destreza en tecnologías y prácticas para poder entregar a tiempo y con calidad. Nos ajustamos a sus necesidades.</p>
```
alvarogarcia7, line 147: Not really sure if 'craftsmen' is understood by Spanish speakers

_i18n/es.yml
```diff
@@ -140,6 +140,47 @@ publications-tweet_feed:
...
+services-software_creation-embedded_craftsmen:
+   title: Craftsmen Embebidos
```
alvarogarcia7, line 149: same here

_i18n/es.yml
CarlosMChica: @alvarogarcia7 I tried to keep consistency and use 'usted'. However, I may have missed some, so please let me know if you spot some other occurrences.

_i18n/es.yml
CarlosMChica: @alvarogarcia7 I also wondered but I think it will be better understood than using something like 'artesano'

CarlosMChica:
services codurance

PR #366, opened 2016-09-14
Events page

ghost:
home page - highlights with videos and events


CarlosMChica:
@halimakoundi what do you think if we change the name of the button "View all blog posts" to "View all publications"? It is now pointing to publications page so I'd say it makes more sense.

CarlosMChica (approved):
Happy to merge it

PR #367, opened 2016-09-14
Add jetbrains partner page spanish translation

CarlosMChica:


dhatanian (changes requested):
This is just a test of the review feature, sorry for the noise

dhatanian (approved):

PR #371, opened 2016-09-15
Show loalised company address in company page map

CarlosMChica:


ghost:
Would have been better to have a different PR to remove travis and the _utility/publish.sh file
But since it is no longer used anyway, it's farly safe to merge

PR #382, opened 2016-09-19
Add padding right to language switcher in small devices

CarlosMChica:


CarlosMChica:
Before:
screen shot 2016-09-19 at 09 58 25

After:
screen shot 2016-09-19 at 09 58 40

PR #398, opened 2016-09-30
adding instructions to readme on publishing videos to the website

ghost:


README.md
```diff
@@ -85,3 +85,10 @@ If you want to push code without deploying a new version of the site, make sure
   * See [Jenkins example post](/site/blob/master/_posts/2014-10-03-guide-to-deploying-artifacts-with-jenkins.md), especially the ``abstract`` field.
   * Note: comments can only be single line
   * Note: Even if they are too long, they'll be cut to 30 words
+
+### Adding a video to the publications
+
+  * Add an md file to the folder: `videos/_posts`
```
jukecraft: maybe add a comment that an existing entry can be used as a template

README.md
```diff
@@ -85,3 +85,10 @@ If you want to push code without deploying a new version of the site, make sure
...
+  * Do not forget the attribute `video-url`, for youtube videos, use following format: https://www.youtube.com/embed/<video-id>
+  * Add an image to the video, for youtube videos, make a screenshot [here](http://youtubescreenshot.com/), the image is used on the home page and the carousel
```
jukecraft: where do we put the image?

README.md
```diff
@@ -85,3 +85,10 @@ If you want to push code without deploying a new version of the site, make sure
...
+
```
jukecraft, line 94: what about the description? Do we take it straight from the YouTube listing?

jukecraft (commented):
questions!

README.md
ghost: I thought it was too obvious :)

README.md
ghost: Well it depends on whatever the publisher feels like doing.. It will be reviewed anyway when PR is submitted I would say.

jukecraft: I meant that there is no guidance that you need one or how to come up with one. But I will merge this anyway. Just something to think about as additional info

PR #427, opened 2016-10-14
Improve 404 page

### LINKED ISSUE #415: Improve 404/403 page
Since we're on AWS S3, any error in the URL results in an ugly 403 access denied page.

We need a proper "not found" page.

Ideally we would have Google Analytics on it to track how many hits we get (and if possible what was the failing URL, maybe in the referrer?).

image

dhatanian:
Fixes #415 

This PR:
- Improves the wording of the 404 page 
- Adds suggestions based on the URL that the user tried to access
- Sends a Google Analytics event indicating the URL the user tried to access and the referrer


404.html
```diff
@@ -3,25 +3,97 @@
...
+  <div class="row">
+    <div class="col-md-8 col-md-offset-2">
+      <h1>Page not found</h1>
+
+      <p>Sorry, we could not find the page you were looking for. You can <a href="/">go to the landing page</a> or use the menu at the top of the page to navigate to a specific area of the site. </p>
+
+      <div id="search-suggestions" style="display:none;">
+        <b>Maybe those suggestions help?</b>
+        <ul id="search-suggestions-list">
+          <li><a href="/">blablablabla</a></li>
```
jukecraft: could this be something more useful? :)

dhatanian:
Without suggestions (when no existing page matches the url):

image

With suggestions:

image

404.html
```diff
@@ -3,25 +3,97 @@
...
-<!-- send a 404 event to google analytics.
-	Source: https://gist.github.com/pwenzel/eb4f04552ef410f60406 -->
-<script>
-if(window.googleAnalyticsEvents){
-    window.googleAnalyticsEvents.functions.push( function(environment) {
-        environment.ga('send', 'event', 'error', '404', 'page: ' + document.location.pathname + document.location.search + ' ref: ' + document.referrer, {'nonInteraction': 1});
-    });
-}
-</script>
+<!-- send a 404 event to google analytics.-->
```
jukecraft: I'll trust you on this

404.html
```diff
@@ -3,25 +3,97 @@
...
+  <script>
+    ga('send', 'event', 'error', '404', 'page: ' + document.location.pathname + document.location.search + ' ref: ' + document.referrer);
+  </script>
+
+  <script type="text/javascript">
```
jukecraft, line 28: since there's no comment here, it looks like the comment above is still valid and this is part of the google analytics event. either move 25 down one line or (preferably) add another comment here

404.html
```diff
@@ -3,25 +3,97 @@
...
+              .attr('href',result.url)
+              .text(result.title)
+            )
+        );
+      });
+      $('#search-suggestions').show();
+    }
+
+    function searchMatchingPages(term){
+      var results = index.search(term)
```
jukecraft, line 48: I'll trust you on this. tested locally with `/client`, worked well.

404.html
dhatanian: Here's how it looks on GA. There is no referrer because I typed the url directly.

image

404.html
dhatanian: Added screenshots in the PR comments

404.html
dhatanian: As discussed, moved to:

image

PR #434, opened 2016-10-17
Change the permalinks in redirects for it to work on AWS S3. Fixes #430

### LINKED ISSUE #430: Missing redirection for /aboutus/ourteam/
Based on Google Analytics events added in #428 , the page `/aboutus/ourteam/` has been accessed twice by users coming from Google, which resulted in a 404. We need to add a redirection to `/company`.

image

dhatanian:


jukecraft:
this probably won't work for e.g. `/aboutus/ourteam`, but google links to `/aboutus/ourteam/` so it should be fine

PR #444, opened 2016-11-02
change author page to `publications` - fixes #443

### LINKED ISSUE #443: publication urls actually contain word "pubication"
https://codurance.com/pubications/author/mashooq-badar/ 200s but  https://codurance.com/publications/author/mashooq-badar/ 404s 😞

jukecraft:
this might break existing author links, but they were already kind of broken once when we introduced the new structure and it didn't seem to cause a lot of trouble. The only way I can think of to fix the problem with old links would be to redirect every single one of them... Let me know if you have a better idea


alvarogarcia7:
@franziskas Does the environment allow for rewrite rules?

jukecraft:
@alvarogarcia7 I have no idea, maybe @halimakoundi knows?

ghost:
I don't think there is an easy way to handle url rewrite from what I remember, we could do something similar to what is done here https://github.com/codurance/site/blob/master/careers/craftsman.html
and redirect anything that goes to pubication to go to publication, so at least we would not have 404. What do you think?
I can do it if you are happy with that.

Sorry for the late reply

CarlosMChica:
@halimakoundi @franziskas is this PR still alive?

ghost:
I added a folder with a redirection file to the pubication page but it does not work for when authors are in the path, I had to make a redirection from the 404 page using javascript.

If everyone is happy with this I think we should merge this PR

PR #456, opened 2016-12-12
Add ALL/POSTS/VIDEOS filters in publications page

CarlosMChica:


CarlosMChica:
videos publications
post publications
all publications

PR #504, opened 2017-01-09
adding some css styling to fix the div height - fixes #500

### LINKED ISSUE #500: Team photos should overflow from the left
The last row in the matrix of photos on the "Company" page puts the new photo on the right, which is a bit unexpected. Please see the bottom of the screenshot. /cc @mashooq  screen shot 2017-01-05 at 17 31 31

ghost:
image


assets/css/custom.css
```diff
@@ -1126,3 +1126,13 @@ h2.gray-underline:after {
...
 }
+
+.row.team div.col-md-2 {
+    height: 55vw;
+}
+
+@media (min-width: 768px){
+ .row.team div.col-md-2{
+    height: 17vw;
+	}
```
neomatrix369, line 1137: minor: closing brace indentation is not in line

neomatrix369:
Layout looks good 👍 and issue is fixed when viewed in normal mode, in all three browsers (Chrome, Firefox, Safari. But when resized to different views, the images inside the page increase in size (sudden jump). This could be due to not all viewports (browser sizes have been configured via css).

Haven;t check on mobile browser - hopefully it will be fine.

assets/css/custom.css
```diff
@@ -1126,3 +1126,13 @@ h2.gray-underline:after {
     -webkit-align-items: center;
     color: #fff;
 }
+
+.row.team div.col-md-2 {
```
neomatrix369, line 1130: Any difference defining a class like:
```
.fixed-image-size {
...
}
```
and using it in the respective places in the HTML rather how it's done here ?

ghost: yeah, probably having a separate class would be better, not sure about the name though

neomatrix369:
@cyplo As agreed, all links are now using `https` protocol, except for one which does not support it.

jukecraft:
@halimakoundi sorry we have conflicts now

ghost:
thanks for letting me know

ghost (commented):
regarding the resize I have added additional media querie to support various viewport. 
so far it is best solution I find within my css skills, happy for any improvement suggestions.

cyplo:
We had a cool chat yesterday with @halimakoundi, we're unsure how to proceed here yet, but it seems that bootstrap is doing their own media queries already to choose different css classes, e.g. `md`, `xs` etc.

ghost:
please check PR #524 for an alternative solution

PR #516, opened 2017-01-10
Fix robots txt typo

PaNaVTEC:
We had a misspell, just fix it. 
This should not affect anything.


neomatrix369:
Would this have an implication (block them?) on scanners like the ones we used to sieve out incorrect links, dead links? /cc @cyplo

PaNaVTEC:
This is just to allow all the content to be indexed. It does not have any side effect. We can remove that line, remove the file or just fix the typo. 

http://www.robotstxt.org/robotstxt.html

cyplo:
I think it should not (i.e. I think this change should be safe). Maybe we could set up an external service that would tell as if the configs allows a known set of robots ? And check periodically ? what do you think ?

neomatrix369:
Okay good idea. The service might also inform us that the service is blocked due to _xyz_ in case it is.

Issue #519, opened 2017-01-11
Migrate Jenkins configuration to be stored in the repo fully


cyplo:
There are some pieces of the configuration, e.g. which scripts to launch and in which order, stored in Jenkins itself. If there are changes done to this configuration it will affect all branches. It would be good imho to have this done per-branch, which new Jenkins can do I believe. Previously it was done by the Pipeline and SCM Sync plugins, I think this is now a part of Jenkins proper.

neomatrix369:
Version control configuration - way to `Infrastructure as code` :)

PaNaVTEC:
This is already done: https://github.com/codurance/site/pull/527

Issue #512, opened 2017-01-10
Add a bot to close stale pull requests


cyplo:
If we would deploy separate sites (#418)  or use other resources per pull-request it would be good to have something automatically close them if they become stale so we can free up those resources.

PaNaVTEC:
This is already deployed into this repository. Closing.

PR #527, opened 2017-01-17
Make Jenkins do a PR preview on a S3 bucket

### LINKED ISSUE #418: Have a preview site with distinct url per pull request

PaNaVTEC:
This is to have a separate site deployment per each PR.
It creates a new S3 bucket per *branch*.
Any subsequent pushes to this branch will result in redeploys, so you can have a fresh site preview any time you want !

Fixes  #418 #519.

How to see the preview:
2017-01-17-132931_1195x638_scrot

As you can see we have now a Jenkins box where it shows the deployment URL, so you can preview the site. Unfortunally for now you have to copy and paste the URL yourself. When we finish the integration with Circle CI it will be fixed as we are going to write a comment instead.

View of the pipeline in the CI:
2017-01-17-132820_964x470_scrot

This PR is to get the preview functionality working, however it does not mean that the work on the CI is done. Next step is to (not done yet)
* Remove the bucket after PR closes 
* Add a bot that auto-closes stale PRs (#512)
* Migrate to CircleCI


neomatrix369:
Good work @PaNaVTEC, I'll add my review comments

_utility/normalize_bucket_name.sh
```diff
@@ -0,0 +1,6 @@
+normalizedBucketName() {
```
neomatrix369: As discussed pls add:
```
#!/bin/bash
set -e
```
to the header of this script

_utility/shorten_url.sh
```diff
@@ -0,0 +1 @@
+curl https://is.gd/create.php\?format\=simple\&url\=$1
```
neomatrix369, line 4: Same here as above, add header suggested above.

_utility/deploy_prb.sh
```diff
@@ -0,0 +1,28 @@
...
+AWS_REGION=$1
+WEBSITE_INDEX_DOCUMENT=$2
+BUCKET_NAME=$3
+
+bucketDoesNotExist() {
+  number_of_buckets_with_name=$(aws s3 ls | awk '{print $3}' | grep -c "^$1$") 
+  echo $(expr $number_of_buckets_with_name = "1") == 0 
+}
+
+BUCKET_NOT_EXISTS=$(bucketDoesNotExist $BUCKET_NAME)
```
neomatrix369, line 22: Minor: why not call the variable: `BUCKET_DOES_NOT_EXIST`

Jenkinsfile
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    stage('Deploy') {
+      BRANCH_NAME=sh(script: 'git describe --all --exact-match 2>/dev/null | sed \'s=.*/==\'', returnStdout: true).trim()
+        AWS_REGION='eu-west-1'
+        WEBSITE_INDEX='index.html'
+        BUCKET_NAME=sh(script: "_utility/normalize_bucket_name.sh site-$BRANCH_NAME", returnStdout: true).trim()
+        sh "_utility/deploy_prb.sh $AWS_REGION $WEBSITE_INDEX $BUCKET_NAME"
+
+        BUCKET_URL="https://${BUCKET_NAME}.s3.amazonaws.com/$WEBSITE_INDEX"
```
neomatrix369: Minor: this is actually the preview site url, how about 'PREVIEW_SITE_URL'

Jenkinsfile
```diff
@@ -0,0 +1,44 @@
...
+        SHORTEN_URL=sh(script: "_utility/shorten_url.sh $BUCKET_URL", returnStdout: true).trim()
+        setBuildStatus("Deployed: $SHORTEN_URL", 'SUCCESS');
+    }
+
+  } catch (e) {
+    setBuildStatus("Error while building", "UNSTABLE");
```
neomatrix369, line 31: Maybe add the contents of the error object to this message or as another message below it? Something like:
```
 setBuildStatus("Error while building. \n Error message: " + e, "UNSTABLE");
```

_utility/deploy_prb.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+set -e
+
+if [ "$#" -ne 3 ]; then
+  echo "Invalid argument count."
+  echo "Usage: $0 AWS_REGION WEBSITE_INDEX_DOCUMENT BUCKET_NAME"
```
neomatrix369: Minor: add an example of the usage

_utility/deploy_prb.sh
```diff
@@ -0,0 +1,28 @@
...
+}
+
+BUCKET_NOT_EXISTS=$(bucketDoesNotExist $BUCKET_NAME)
+if [ $BUCKET_NOT_EXISTS ]; then
+  aws s3 mb s3://$BUCKET_NAME --region $AWS_REGION
+  aws s3 website s3://$BUCKET_NAME --index-document $WEBSITE_INDEX_DOCUMENT	
+fi
+
+aws s3 sync . s3://$BUCKET_NAME --acl public-read
+aws cloudfront create-invalidation --distribution-id EML2BUMD54HSC --paths "/*"
```
neomatrix369, line 29: We need to find out if this distribution id `EML2BUMD54HSC` can be reused for our jobs, or do we need a new one, maybe @mashooq can recommend us something

Jenkinsfile
PaNaVTEC: Sure, much better. thanks for the suggestion

Jenkinsfile
PaNaVTEC: Not sure about that, the limit for that string is 140 characters, And it will be displayed in the box that you can see at the screenshot, so it does not seem really usable.

_utility/deploy_prb.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+set -e
+
+if [ "$#" -ne 3 ]; then
+  echo "Invalid argument count."
```
neomatrix369: I would say something like this: 
"Insufficient arguments, you have only provided: "$@", see usage and example for further assistance.".

_utility/normalize_bucket_name.sh
neomatrix369: Might not be necessary to encapsulate into a separate (single) function inside the body of this script, but I'll leave this for you to decide.

cyplo:
It seems that we lost the status for this particular PR. Would it be possible to arrange having this build as the only build for this PR to test it ?

PaNaVTEC:
@neomatrix369 Thanks for the review, all your comments have been addressed. 
@cyplo I forgot to disable the hook in the other repository, so, I think it was me pushing some commits in the other repo, I just disabled it, let's see if it works with both plugins at the same time.

Issue #545, opened 2017-01-23
Cannot access article in the site preview


cyplo:
https://site-remove-author-twitter.s3-eu-west-1.amazonaws.com/2017/01/16/hierarchy-of-opinion/

gives me 
![image](https://cloud.githubusercontent.com/assets/217899/22213540/7bb78adc-e18c-11e6-95c2-58ca05897c82.png)

PaNaVTEC:
The problem seems to be something related with the configuration of the bucket. if you click in this link instead: [https://site-remove-author-twitter.s3-eu-west-1.amazonaws.com/2017/01/16/hierarchy-of-opinion/index.html](https://site-remove-author-twitter.s3-eu-west-1.amazonaws.com/2017/01/16/hierarchy-of-opinion/index.html) it works.
Since we are trying to move towards the approach of having one bucket instead of creating one per pr, I will deffinitly address this, but after addressing this issue: https://github.com/codurance/site/issues/512

PaNaVTEC:
This is already fixed, the problem was that we dind't have a root index.html in the bucket. After putting a random index.html in the root of the site-pr bucket it started to work.

cyplo:
I think that for Sandro's recent PR, when I clicked around I got the access denied XML message again.

PaNaVTEC:
Weird, I've seen it working when @mashooq created an index.html in the root of the bucket.

cyplo:
Maybe it would be worth it to write a simple automated check for that, e.g. using curl, that would run after deployment ?

lpalma:
Is this issue still relevant? I haven't heard of/had problems visualizing the site preview so far.

PaNaVTEC:
I think is already fixed

PR #550, opened 2017-01-27
Preview PRS in one bucket with different folders

PaNaVTEC:
Previously https://github.com/codurance/site/pull/527 we deployed every PR to one new s3 bucket. In this PR we are switching to deploying it in a folder in an already created bucket.

The benefits of this approach are:
- Security, because we can strip the bucket deletion permission from the process.
- Better control over what goes where.

This will help us to autommatically delete the site preview when any PR is merged.


_utility/deploy_prb.sh
```diff
@@ -1,32 +1,16 @@
...
-if [ $BUCKET_NOT_EXISTS ]; then
-  echo "No previous bucket with name $BUCKET_NAME found, creating a new one"
-  aws s3 mb s3://$BUCKET_NAME --region $AWS_REGION
-  aws s3 website s3://$BUCKET_NAME --index-document $WEBSITE_INDEX_DOCUMENT	
-fi
-
-echo "Synchronizing bucket: $BUCKET_NAME"
-aws s3 sync . s3://$BUCKET_NAME --acl public-read
-aws cloudfront create-invalidation --distribution-id EML2BUMD54HSC --paths "/*"
+echo "Synchronizing to folder: $FOLDER_NAME"
```
neomatrix369, line 15: Where is `FOLDER_NAME` getting set? `$1` may not be applied to `FOLDER_NAME`

_utility/send_gh_comment.sh
```diff
@@ -12,9 +12,9 @@ REPO_OWNER=$1
 REPO_NAME=$2
 PR_NUMBER=$3
 AUTH_TOKEN=$4
-BUCKET_NAME=$(_utility/normalize_bucket_name.sh $5)
+DEPLOYMENT_URL=$5
 
-COMMENT="Deployed: $BUCKET_NAME"
+COMMENT="Deployed: $DEPLOYMENT_URL"
 LAST_COMMENT_ID=$(curl https://api.github.com/repos/$REPO_OWNER/$REPO_NAME/issues/$PR_NUMBER/comments?access_token=$AUTH_TOKEN -X GET | jq '[.[] | select(.user.login=="CoduranceBot")][0].id')
```
neomatrix369: Minor: either format the line by using `\` to split the lines or use functions for each separate functionality - will help readability

_utility/deploy_prb.sh
neomatrix369: Ignore me, `FOLDER_NAME` is a CircleCI env variable

circle.yml
```diff
@@ -3,10 +3,10 @@ machine:
     version: 2.3.1
   environment:
     PR_NUMBER: $(echo ${CI_PULL_REQUEST##*/})
-    BUCKET_NAME: "site-$CIRCLE_BRANCH"
+    FOLDER_NAME: "site-$CIRCLE_BRANCH"
     WEBSITE_INDEX_DOCUMENT: index.html
     AWS_REGION: eu-west-1
-    PREVIEW_URL: "https://${BUCKET_NAME}.s3.amazonaws.com/$WEBSITE_INDEX_DOCUMENT"
+    PREVIEW_URL: "https://codurance-site-pr.s3.amazonaws.com/$FOLDER_NAME/$WEBSITE_INDEX_DOCUMENT"
```
neomatrix369: why not make `codurance-site-pr` into an env variable, just like $FOLDER_NAME, helps see the constant parts in the URL from the variable/substitution parts

_utility/deploy_prb.sh
```diff
@@ -1,32 +1,16 @@
...
+aws s3 sync . s3://codurance-site-pr/$FOLDER_NAME --acl public-read
+aws cloudfront create-invalidation --distribution-id EML2BUMD54HSC --paths "/$FOLDER_NAME/*"
```
neomatrix369, line 17: Check if we need this?

neomatrix369: Also, any other setting to make site crawlable like production site, referring to issue #545

_utility/send_gh_comment.sh
neomatrix369: "CoduranceBot" ==> make it a constant/variable

neomatrix369:
@PaNaVTEC Had a quick glance, that's all I can see, happy to discuss any point you like to

_utility/deploy_prb.sh
PaNaVTEC: Yes but you are right, we are passign it inside the script so it should be the case. Thanks!

_utility/deploy_prb.sh
PaNaVTEC: Yes, we need this line for now, that invalidates the cdn cache if not it will not be correctly displayed. I think the Issue that you linked is not related, that was related with permissions or the index document option that you can specify.

PaNaVTEC:
@neomatrix369 addressed all your comments, let me know if there is anything else 👍

_utility/deploy_prb.sh
neomatrix369: I have changed the issue id since then, thanks for the correct reference.

neomatrix369:
I think this PR is good to merge, @cyplo @mashooq do you have any comments for @PaNaVTEC

cyplo:
looks good to me

PR #551, opened 2017-01-27
Fix image width because of class row blog page img

ananogal:
Please be aware that images should be used with the `` tag and have applied to it the class "img-responsive". This will solve the problem of rendering big size images in mobiles.

### DIFF
--- a/_posts/2015-03-16-installing-zprezto-a-quick-guide.md
+++ b/_posts/2015-03-16-installing-zprezto-a-quick-guide.md
@@ -37,7 +37,7 @@ Zsh offers a plethora of additional features on top of bash that you may find a 
 
 Some features of note that I found particularly useful when I first made the switch are:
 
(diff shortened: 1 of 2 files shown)


cyplo:
Would it mean we need to change the definitions for all the old posts containing images ? If so - what do you think about making this transformation be a part of the build process instead of changing them manually in the .md files ?

cyplo:
Had a chat with @ananogal  in the meatspace - thanks a lot for that :)
To summarise it here:
I think this is okay to merge, as there are other posts that do the same magic with tags.

I assumed, wrongly, that most people use the markdown tags for posts by default and that is the most natural thing to do when doing the drive-by contributions, i.e. people not very familiar with the codebase.

We did a quick grep and the image insertion solutions are all over the place - we need a separate discussion on the conventions here.

PR #554, opened 2017-01-30
Fix CI fail when sending GH comment

### LINKED ISSUE #546: Failure to send a comment fails the build
As in a build for  #542 

would it be okay to allow failures in this section of the build ? what do you think ?

PaNaVTEC:
This fixes #546 

What was happening is that we were opening pull request while the CI was building the last commit  that we pushed.

I modified the script to:
- Don't fail the build if the github comment can't be sent.
- If the CI is building and we don't have a PR but at the time that we are about to send the comment we have created the PR, the modification on the script will get the PR number and the comment will be send. This is done triggering a github api call with the branch name if we don't have the PR number when we are about to send the comment.
- Check arguments and set names like `-o`. I've used a generator that you can find here: http://getoptgenerator.dafuer.es/


_utility/send_gh_comment_validation.sh
```diff
@@ -0,0 +1,105 @@
+# Define help function
```
cyplo, line 4: A note on how did you generate this and how a next maintainer might re-generate with minimal effort might be helpful here

_utility/send_gh_comment.sh
```diff
@@ -1,34 +1,43 @@
...
-  echo -e "Usage: $0 REPO_OWNER REPO_NAME PR_NUMBER AUTH_TOKEN DEPLOYMENT_URL\n"
-  echo "Example: $0 codurance site 4 token https://bucket-name.amazon.com/index.html"
-  exit 1
+REPO_OWNER=''
+REPO_NAME=''
+PR_NUMBER=''
+GITHUB_TOKEN=''
+DEPLOYMENT_URL=''
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
neomatrix369: would a combination of ' and " on this line work - would make it easier to read and understand whats happening

_utility/send_gh_comment.sh
```diff
@@ -1,34 +1,43 @@
 #!/bin/bash
-set -e
+set +e
```
neomatrix369: Either create a usage() that displays the usage with an example or put that text in the header of this file as comments whichever is more appropriate - will help with the learning curve.

_utility/send_gh_comment.sh
```diff
@@ -1,34 +1,43 @@
...
+REPO_OWNER=''
+REPO_NAME=''
+PR_NUMBER=''
+GITHUB_TOKEN=''
+DEPLOYMENT_URL=''
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "$DIR/send_gh_comment_validation.sh"
+
+if [ -z "$PR_NUMBER" ]
```
neomatrix369: Maybe extract this into a function of its own, so the main body is more readable.

_utility/send_gh_comment.sh
```diff
@@ -1,34 +1,43 @@
...
+	fi
 fi
 
-GITHUB_USERNAME='CoduranceBot'
-REPO_OWNER=$1
-REPO_NAME=$2
-PR_NUMBER=$3
-AUTH_TOKEN=$4
-DEPLOYMENT_URL=$5
+LAST_COMMENT_ID=$(curl https://api.github.com/repos/$REPO_OWNER/$REPO_NAME/issues/$PR_NUMBER/comments?access_token=$GITHUB_TOKEN \
```
neomatrix369: Just like above, can be extracted into a function

_utility/send_gh_comment.sh
```diff
@@ -1,34 +1,43 @@
...
+if [ $? -ne 0 ]; then
+  echo "Error while parsing last commend id"
+  exit 1
+fi
 
 COMMENT="Deployed: $DEPLOYMENT_URL"
-LAST_COMMENT_ID=$(curl https://api.github.com/repos/$REPO_OWNER/$REPO_NAME/issues/$PR_NUMBER/comments?access_token=$AUTH_TOKEN \
-  -X GET | jq "[.[] | select(.user.login==\"$GITHUB_USERNAME\")][0].id")
-
 if [ "$LAST_COMMENT_ID" != "null" ]; then
```
neomatrix369: Extract into function

_utility/send_gh_comment_validation.sh
```diff
@@ -0,0 +1,105 @@
...
+function help(){
```
neomatrix369, line 5: The functions that do not belong directly to this file can be placed elsewhere (another file) and we can source it in, so they are still visible to this file - we can pair on this if you like.

_utility/send_gh_comment.sh
PaNaVTEC: That is done in "send_gh_comment_validation.sh" but we splitted this in 2 scripts due to its verbosity 👍

_utility/send_gh_comment_validation.sh
PaNaVTEC: This file has been autogenerated, I rather prefer leave it like it is, so if someone needs to add a new parameter in the future its as easy as autogenerate it in the webpage that I wrote in the comment in the header description

PaNaVTEC:
@cyplo @neomatrix369 I addressed all your comments, if you are happy with the PR, :shipit:

_utility/send_gh_comment_validation.sh
neomatrix369: I think you just did what I suggested so that cool.

neomatrix369:
It looks good to me, @cyplo you have any comments?

Issue #529, opened 2017-01-18
Tidying up and refactoring custom.css


neomatrix369:
The `custom.css` file in the project folder under `/assets/css/` isn't organised or individual CSS configurations are NOT placed in a manner we can find them easily and know which category they belong to. A good grouping/categorisation could be something like this:

- Layout
- Typography
- Navbar
- Carousel
- Global Styles
- Media Queries
  - Portrait & landscape phone
  - Landscape phone to portrait tablet
  - Large desktop

So we could either re-organise CSS config that apply to the above categories in one file or split the `custom.css` files into multiple files, one file per category (i.e. `custom-layout.css`, etc).

**Bonus points: ** the configuration are ordered in increasing order of their specificity (see [examples of specificity](https://www.w3.org/TR/css3-selectors/#specificity)). Lower the specificity, more general or generic is the CSS configuration, on the other hand, a high number is a sign of a very specific CSS configuration.

Some examples of the above categories:
- Layout
```css
#features {
   text-align: center;
}
```

- Typography
```css
body {
   -webkit-font-smoothing: antialiased;
   text-rendering: optimizeeligibility;
}
```

- Navbar
```css
#searchInput {
   width: 200px;
}
```

- Carousel
```css
#myCarousel {
   margin-top: 60px;
}
```

- Global Styles
```css
.alert {
   margin-top: 20px;
   display: none;
}
```

- Media Queries
     - Portrait & landscape phone
```css
@media (max-width: 480px) {
   .thumbnail .price {
      font-size: 14px;
   }

   .thumbnail .price sup {
      font-size: 10px;
   }
}
```
- Media Queries
     - Landscape phone to portrait tablet
```css
@media (max-width: 768px) {
   #slide1 {
      background: url('../../images/carousel_small_01.jpg') top center no-repeat;
   }

   #slide2 {
      background: url('../../images/carousel_small_02.jpg') top center no-repeat;
   }

   #slide3 {
      background: url('../../images/carousel_small_03.jpg') top center no-repeat;
   }	

   .carousel-caption {
      font-size: 16px;
   }

   .carousel-caption h4 {
      font-size: 24px;
   }
}
```
- Media Queries
     - Large desktop
```css
@media (min-width: 1200px) {
   #slide1 {
      background: url('../../images/carousel_large_01.jpg') top center no-repeat;
   }

   #slide2 {
      background: url('../../images/carousel_large_02.jpg') top center no-repeat;
   }

   #slide3 {
[...]

PR #561, opened 2017-02-07
Refactor custom css file buttons

neomatrix369:
- Removed unused button CSS selector and references to it in the rest of the CSS file
- Moved the buttons and grouped them together.  …
Added CSS specificity for each of them and sorted them in order of specificity (smaller to the larger, more generic to more specific)
- Added the missing selector, moved it to the top of the list after generating a CSS specificity for it

Related to issue #529


neomatrix369:
Reviewed the relevant pages and happy to take comments.

PR #565, opened 2017-02-08
Refactor custom css file add css speficity to selectors

neomatrix369:
PLEASE MERGE THIS PR FIRST

Related to issue #529

PR #567, opened 2017-02-09
Refactor custom css file split into multiple files

neomatrix369:
PLEASE ONLY MERGE THIS AFTER MERGING PR: https://github.com/codurance/site/pull/565

- Extracted global styles, layout, navbar/navigation, buttons, colours and media queries from the custom.css into their own files (custom-xxxx-xxxxx.css) and added them to the head.html to make them available to all other pages
- Added CSS specificity to the media queries and a link to the resource on how to calculate it
- Reordered a few selectors in the order of specificity - smallest to highest (more generic to more specific)
- Renamed custom.css to custom-layout.css and also changed it in the head.html to make it available to all other pages

Related to issue #529


jukecraft:
Hi Mani,

Is this one on top of #565? Hard to see what changes are only in this one

PR #568, opened 2017-02-09
Refactor custom css file split into multiple files

neomatrix369:
- Extracted global styles, layout, navbar/navigation, buttons, colours and media queries from the custom.css into their own files (custom-xxxx-xxxxx.css) and added them to the head.html to make them available to all other pages
- Added CSS specificity to the media queries and a link to the resource on how to calculate it
- Reordered a few selectors in the order of specificity - smallest to highest (more generic to more specific)
- Renamed custom.css to custom-layout.css and also changed it in the head.html to make it available to all other pages
- Amended scripts and config files related to building the Jekyll site on the localmachine

Related to issue #529


Rakefile
```diff
@@ -7,12 +7,8 @@ RSpec::Core::RakeTask.new(:spec) do |task|
   task.rspec_opts = '--color'
 end
 
-task :build do
-  sh 'bundle exec jekyll build --config _config.yml,_config_en.yml --trace'
-end
-
 task :buildboth do
-  sh 'bundle exec jekyll build --trace'
+  sh 'bundle exec jekyll build --config _config.yml,_config_en.yml --trace'
```
jukecraft: why don't we want to have a default build option?

assets/css/custom/custom-layout-buttons.css
```diff
@@ -0,0 +1,86 @@
...
+    background-color: transparent;
+    border: 1px solid white;
+    color: white;
+}
+
+.btn-u-orange {   /* 0 0 1 0 */
+    background: #e67e22;
+}
+
+#sharethisButtons {   /* 0 0 1 0 */
```
jukecraft: why is this one not using hyphens?

assets/css/custom/custom-layout-buttons.css
```diff
@@ -0,0 +1,86 @@
...
+.btn-brd.btn-u-orange:hover {   /* 0 0 3 0 */
+    color: #d35400;
+    border-color: #d35400;
+}
+
+.btn-brd.btn-u-orange.btn-brd-hover:hover {   /* 0 0 4 0 */
+    background: #d35400;
+}
+
+#sharethisButtons {   /* 0 1 0 0 */
```
jukecraft, line 80: duplicate from above?

assets/css/custom/custom-layout-colours.css
```diff
@@ -0,0 +1,71 @@
...
+
+.white,
+p.white,
+ul.white li {
+    color: white;
+}
+
+/* Section colours */
+
+.gray-section {   /* 0 0 1 0 */
```
jukecraft: duplicate?

assets/css/custom/custom-navbar-or-navigation.css
```diff
@@ -0,0 +1,36 @@
...
+   Navbar or Navigation
+   -------------------------------------- */
+
+/*
+   CSS specificity calculator: https://specificity.keegan.st/
+   Specificity components: [Inline styles] [IDs] [Classes, attributes and pseudo-classes] [Elements and pseudo-elements]
+   Specificity ranges from lowest to highest (more generic to more specific)
+ */
+
+.breadcrumbs h2 {  /* 0 0 1 1 */
```
jukecraft: is this part of the navigation?

buildAndRunSite.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-jekyll serve --incremental --watch
\ No newline at end of file
+jekyll serve --incremental --watch --config _config.yml,_config_en.yml
```
jukecraft, line 3: maybe change the README as well?

Rakefile
neomatrix369: It's doable, and we intend to do it cleaner in another PR, but that would mean changes to the config files (both Rake and _config files). 

This change here was to remove tasks that didn't work or was redundant.

buildAndRunSite.sh
neomatrix369: The readme refers to running the buildAndRunSite.sh, but we will be changing the way we build the two sites, so not yet.

assets/css/custom/custom-layout-buttons.css
neomatrix369: indeed, i;'ll remove it

assets/css/custom/custom-navbar-or-navigation.css
neomatrix369: Its not used, i'll remove it - thnx

neomatrix369:
Checked the relevant pages, no regression found after applying these changes.

Rakefile
jukecraft: For now, can we not keep the build task even if it's a copy?

jukecraft:
@neomatrix369 can you take care of the conflicts please?

assets/css/custom/custom-layout-buttons.css
jukecraft: I think you  removed the other one, the name still stands out

neomatrix369:
@franziskas just resolved the conflicts

Rakefile
neomatrix369: the old build task won't work without the added parameters, which makes one of them redundant

assets/css/custom/custom-layout-buttons.css
neomatrix369: Not changing names of selectors via this PR we could do this via another one

assets/css/custom/custom-layout-buttons.css
neomatrix369: As mentioned earlier not changing the name of this selector, in this PR, maybe via another PR

assets/css/custom/custom-layout-colours.css
neomatrix369: Has been removed

Rakefile
jukecraft: Exactly. Why not keep the one from the readme people are used to?  You're breaking the interface

neomatrix369: Just added a commit to address this so the README instructions stay good

PR #572, opened 2017-02-14
Convert css to scss custom layout colours

neomatrix369:


assets/css/custom/custom-layout-colours.scss
```diff
@@ -10,10 +14,12 @@
 
 /* Section colours */
 
+$solid-one-px-thick-yellow: solid 1px #eee;
```
PaNaVTEC, line 17: The color name seems to be grey

PR #573, opened 2017-02-14
Convert css to scss custom layout buttons

neomatrix369:

Refactoring involved:

- simplification of the selectors - bringing them to the unexpanded form
  - grouping by their specificity
  - grouping by elements and pseudo-elements
- extracted a commonly used colour across multiple selectors

### DIFF
--- a/assets/css/custom/custom-layout-buttons.scss
+++ b/assets/css/custom/custom-layout-buttons.scss
@@ -1,3 +1,7 @@
+---
+# Front matter comment to ensure Jekyll properly reads file.
+---
+
 /* --------------------------------------
    Layout
    -------------------------------------- */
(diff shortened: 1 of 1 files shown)


assets/css/custom/custom-layout-buttons.scss
```diff
@@ -10,6 +14,11 @@
 
 /* Buttons - used across multiple pages */
 
+$dark-orange-colour: #d35400;
```
PaNaVTEC, line 17: I think it could be a good idea to extract all the colors to one file, and we can use it anywhere. In Android I used to have a `colors-global`, `colors-screen-main`, `colors-screen-X/Y`. With that you can have all the colors together, and is easier to reuse instead of declare new ones.

neomatrix369: 👍 Great point, as these elements become more and more visible I'll be extracting them into separate file(s), importing them and creating new PRs - thanks @PaNaVTEC

PR #574, opened 2017-02-14
Convert css to scss custom layout

neomatrix369:

Refactoring involved:

- simplification of the selectors - bringing them to the unexpanded form
  - grouping by their specificity
  - grouping by elements and pseudo-elements
- extracted a commonly used colour across multiple selectors
- reorganised selectors to bring similar group of selectors together


neomatrix369:
Compared old and new CSS using http://www.alanhart.co.uk/tools/compare-css.php#CSSText - they are the same now, even though the order of the selectors have been changed in the new version.

PR #577, opened 2017-02-15
Specify the default language (i.e. EN) when building the site using the build bash script

neomatrix369:
Passing the _config_default.yml file to the jekyll command-line to specify a default language when building the site, using the build bash script


buildAndRunSite.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-jekyll serve --incremental --watch --config _config.yml,_config_en.yml 
\ No newline at end of file
+jekyll serve --incremental --watch --config _config_default.yml,_config_en.yml
```
PaNaVTEC, line 3: Hey mani, I think what you need here is `--config _config.yml _config_default.yml` That will tell jekyll to serve only the english the language when using this script. How is it now, I am not sure if it will work, _config_default has only the language specification

PaNaVTEC (changes requested):

buildAndRunSite.sh
neomatrix369: will this end up generating the site for both `en` and `es`, and make `en` default - I thought we had agreed upon using ` _config_default.yml,_config_en.yml`, which does generate both `en` and `es`?

neomatrix369: Just ran it and it does not generate `es`, but makes `en` default - which the previous option did

PR #579, opened 2017-02-16
Fix timeouts when PR has not opened yet

PaNaVTEC:
We were using a bash argument generator to validate the `send_gh_comment.sh` script. The problem was that the validator is checking if the arguments are bound to a value with the `-n` condition. But in case that we are sending an empty value we don't shift the parameters anymore. 

That only happened when you push to a branch and you still don't have created the PR. With this fix, we shift the arguments if they are empty, avoiding infinite loops.


_utility/send_gh_comment.sh
```diff
@@ -15,44 +15,49 @@ source "$DIR/send_gh_comment_validation.sh"
...
 
 function get_pr_number_by_branch_name() {
-  PR_NUMBER=$(curl -X GET -u ${GITHUB_TOKEN}:x-oauth-basic "https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/pulls?head=${REPO_OWNER}:${CIRCLE_BRANCH}" | jq ".[0].number")
-	if [ -z "$PR_NUMBER" ]; then
-		echo "PR_NUMBER not specified, and can't be determined in the current context"
-		exit 1
-	fi
+  if [ -n "$CIRCLE_BRANCH" ]; then
+    pr_number_by_branch=$(curl -X GET -u ${GITHUB_TOKEN}:x-oauth-basic "https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/pulls?head=${REPO_OWNER}:${CIRCLE_BRANCH}" | jq ".[0].number")
+    if [ $? -ne 0 ] && [ "$pr_number_by_branch" != "null" ]; then
```
neomatrix369: when or by which action is `pr_number_by_branch` set to `null`? `jq` ?

PaNaVTEC: yep, jq returns null if is not able to parse it

neomatrix369:
`_utility/send_gh_comment_validation.sh`  - have you tried passing various (combinations of)  parameters to this script to see if the `shift` is helping - simulating a few cases cause by  CircleCI/Github might help.

_utility/send_gh_comment.sh
```diff
@@ -15,44 +15,49 @@ source "$DIR/send_gh_comment_validation.sh"
...
+    fi
+  fi
 }
 
 function get_last_comment_id_generated_by_bot() {
   LAST_COMMENT_ID=$(curl https://api.github.com/repos/$REPO_OWNER/$REPO_NAME/issues/$PR_NUMBER/comments?access_token=$GITHUB_TOKEN -X GET | jq "[.[] | select(.user.login==\"$GITHUB_USERNAME\")][0].id")
   if [ $? -ne 0 ]; then
-    echo "Error while parsing last commend id"
-    exit 1
+    echo "null"
```
neomatrix369: I would extract this into a well-name constant - "null", easy to miss out during future updates

_utility/send_gh_comment.sh
```diff
@@ -15,44 +15,49 @@ source "$DIR/send_gh_comment_validation.sh"
...
 function create_comment() {
-	echo "Creating a comment with deployment url"
-	if [ "$EXECUTE" == "1" ]; then
-		curl https://api.github.com/repos/$REPO_OWNER/$REPO_NAME/issues/$PR_NUMBER/comments?access_token=$GITHUB_TOKEN \
-			-H "Content-Type: application/json" \
-			-X POST \
-			-d "{ \"body\":\"$COMMENT\" }"
-	fi
+  echo "Creating a comment with deployment url"
+  if [ "$EXECUTE" == "1" ]; then
```
neomatrix369: minor: "1" what does it stand for? maybe a constant

neomatrix369:
👍  the echos are useful - prints human readable info

_utility/send_gh_comment.sh
```diff
@@ -15,44 +15,49 @@ source "$DIR/send_gh_comment_validation.sh"
...
-	if [ "$EXECUTE" == "1" ]; then
-		curl https://api.github.com/repos/$REPO_OWNER/$REPO_NAME/issues/comments/$LAST_COMMENT_ID?access_token=$GITHUB_TOKEN \
-			-H "Content-Type: application/json" \
-			-X POST \
-			-d "{ \"body\":\"$COMMENT\" }"
-	fi
+  echo "Found a previous comment with id:$LAST_COMMENT_ID . Updating comment"
+  if [ "$EXECUTE" == "1" ]; then
+    curl https://api.github.com/repos/$REPO_OWNER/$REPO_NAME/issues/comments/$LAST_COMMENT_ID?access_token=$GITHUB_TOKEN \
+      -H "Content-Type: application/json" \
```
neomatrix369: minor: make it a constant and replace repeats with it

PaNaVTEC:
Yep I've been able to reproduce the issue in my machine by sending to this script empty strings as values for the named arguments. I tried with different combinations and seems to work. The problem was the `while true` in combination of not shift the arguments.

_utility/send_gh_comment.sh
neomatrix369: Can there be a case where `$pr_number_by_branch` does not return a "null"?

PaNaVTEC: if the exit status is 0 there shouldn't be a case. If the exit status is 1, then yes.

neomatrix369:
@PaNaVTEC finished reviewing, see if the other comments are worth applying, otherwise, the PR is good to be merged.

_utility/send_gh_comment.sh
PaNaVTEC: I don't know, in bash is pretty standard to do this because there are no booleans.

PaNaVTEC:
@neomatrix369 I've addressed your comments, let me know what you think

PR #582, opened 2017-02-22
Update site template to 1.9.8

PaNaVTEC:
Update template to 1.9.8:
- Removed jekyll-less since we don't use it
- @neomatrix369 and me noticed a change in the hover effect in the company page. That seems to be an update in the style
- Moved some styles from blocks.css (file belonging to the template) to custom/custom-blocks.css (our own customization)

We've run the site-diff-tool locally, and we've didn't noticed any breaking changes


PaNaVTEC:
site preview: https://codurance-site-pr.s3.amazonaws.com/site-update_site_template/index.html

PR #601, opened 2017-03-23
Adding robots.txt to make the sitemap visable to google for codurance…

sjn88:
….com

This doesn't make the sitemap work for codurance.es (broken on master) nor does it make robots.txt on codurance.es work. These will be in a future PR.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -40,9 +40,9 @@ pagination:
 
 gems:
   - jekyll-redirect-from
-  - jekyll-sitemap
   - jekyll-paginate-v2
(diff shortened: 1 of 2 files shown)


sjn88:
I've added an extra commit that may fix robots.txt and the sitemap if the url is set to the right localised url, otherwise it will fail.

sjn88:
At the very least it uses the default site url. https://codurance-site-pr.s3.amazonaws.com/site-sitemapToRobots.txt/sitemap.xml *crosses fingers for codurance.es*

PR #610, opened 2017-03-30
move contact us one level up

MissArmieJ:


sjn88:
• "contact us" should be orange instead of grey
• the hover of "contact us" should be white instead of grey.
• The links to the Spanish and English website version are still too close to the "Contact us" button. They should be aligned to the "Y" of "COMPANY""

sjn88:
I am unhappy with the space between the button and the nav bar below, but I think we need to bring this in to close off a bug https://github.com/codurance/site/issues/613

PR #620, opened 2017-04-11
Fixed positioning nav bar

charlyraffellini:


MissArmieJ (approved):
Giulia is happy and confirmed the same with a few other people in the office (no complaints from the Slack channel either) - so I'll Approve and merge.

PR #622, opened 2017-04-12
updates for virtualbox install w Robert

MissArmieJ:


Gemfile.lock
```diff
@@ -220,4 +220,4 @@ DEPENDENCIES
   travis
 
 BUNDLED WITH
-   1.14.5
```
robertfirek, line 223: This change is not needed.

PR #625, opened 2017-04-24
Armanda's first blog - on context

MissArmieJ:
Could you all please review my blog? Your comments will be much appreciated :-)


_i18n/en.yml
```diff
@@ -129,7 +129,7 @@ company-contact:
    title: Contact us
 company-contact-form:
    title: Get In Touch
-   subtitle: Call us on <a href="tel:+442074902967">+44 207 4902967</a>, email us at <a href="mailto:hello@codurance.com">hello@codurance.com</a>, or use the form below to send us a message.
+   subtitle: Call us on +44 207 4902967, email us at <a href="mailto:hello@codurance.com">hello@codurance.com</a>, or use the form below to send us a message.
```
srodrigo: This change, and similar ones below, look like something unrelated to your blogpost. Is there a reason for having both things in the same PR?

docker/Dockerfile-local
```diff
@@ -1,15 +1,19 @@
 FROM ruby:2.3.1
 
+RUN apt-get update && \
```
srodrigo: I haven't run the website locally in a while. Are these changes, involving Docker, related to the blog post you are creating, or could they be part of a different PR?

srodrigo (commented):
Hi Armanda,
I had a brief look at your PR, and looks to me like there are 3 different things:
- your blog post - this is the main reason of the PR, as far as I understand
- a few changes to the website (telephone numbers)
- a few changes around Docker and the way we run the website locally

Maybe, for next time, it could be split in different PRs.

_i18n/en.yml
MissArmieJ: I may have merged in changes from a different branch.. I'll look into this - thanks

docker/Dockerfile-local
MissArmieJ: There is a different PR for this called "docker-changes" - think it may still be under review. Cheers

PR #659, opened 2017-05-22
Fix problem with missing dependency when run from local machine

robertfirek:


robertfirek:
@TobyRet I think this should solve the problems with your local Ruby error during starting.

PR #667, opened 2017-06-02
new post + user name correction

guillemf:


etorres (commented):
This is looking great! I have a few comments about the text:
* "branching is cheap in Git, so we shouldn't be afraid of branching and testing", maybe branching and merging?
* "There is a great post about how to write a good commit message [here].", link is missing.
* "To do that you have two great tools to the rescue [fastlane] and [Fabric]", include a link to each tool?
* "```fastlane feature", "``fastlane beta": format is missing.

Also, it would be great if the text provides the reader with a brief introduction to the different QA roles of the team. For example, it's not clear to me what differences there are between a tester and an external tester.

guillemf:
@etorres Thanks for your comments, fixed the errors :D ... I am saving your suggestions for the next post.

PR #692, opened 2017-06-26
Improve quality of the page for Open Graph Protocol parsers

robertfirek:
Probably Linkedin parser has some problems with the missing space at the end of the `meta` tag. I also fixed some minor bugs in our page template which may cause some problems for parsers.


cyplo:
Checked the build's preview , looks more sharp to me, I like it :)

PR #695, opened 2017-06-26
New Craftsman Page

lpalma:
### Not ready for merge

Continuing the branch created by @mashooq, this PR:

- [x] Replace the `Apply` buttons on the Careers page with links to the Apprentice and Craftsman pages;
- [x] Add the Application Form buttons to Apprentice / Craftsman pages;
- [x] Add the proper content to the Apprentice page;

The Application Form used in this PR is the same as the currently live version. The new Application Form will be implemented in a future iteration.

PR #697, opened 2017-06-30
Revert "New Craftsman Page"

mashooq:
Reverts codurance/site#695


robertfirek:
Why do you revert it?

PR #700, opened 2017-07-03
Add mandatory

MiyamotoAkira:
6 sections on the form are now mandatory.

### DIFF
--- a/_includes/apply_form.html
+++ b/_includes/apply_form.html
@@ -65,7 +65,8 @@
                     <div class="row">
                         <div class="col-md-12 ">
                             <div id="input-error-msg-{{ include.target }}" class="alert alert-danger fade in hide">
-                                <strong>Please: </strong> enter all the requested fields.
+                                <div id="fields">
+                                </div>
                             </div>
                         </div>
                     </div>
--- a/_includes/apply_form.html
+++ b/_includes/apply_form.html
@@ -10,18 +10,18 @@
                     <div id="{{ include.target }}-fields">
                         <div class="row">
                             <div class="col-md-6 margin-bottom-20">
-                                <label>First Name</label><input name="{{ include.target }}-firstname" class="form-control" type="text" />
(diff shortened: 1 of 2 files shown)


_includes/apply_form.html
```diff
@@ -10,18 +10,18 @@ <h4 class="modal-title" id="myModalLabel4">{{ include.title }}</h4>
                     <div id="{{ include.target }}-fields">
                         <div class="row">
                             <div class="col-md-6 margin-bottom-20">
-                                <label>First Name</label><input name="{{ include.target }}-firstname" class="form-control" type="text" />
+                                <label>First Name (*)</label><input name="{{ include.target }}-firstname" class="form-control" type="text" />
```
robertfirek, line 13: You can add required option to the field (e.g. `<input type="text" name="..." required>`)

_includes/apply_form.html
```diff
@@ -51,21 +56,22 @@ <h4 class="modal-title" id="myModalLabel4">{{ include.title }}</h4>
                         </div>
                         <div class ="row">
                             <div class="col-md-12 margin-bottom-20">
-                                <label>What does Software Craftmanship mean to you?</label>
+                                <label>What does Software Craftmanship mean to you? (*)</label>
                                 <textarea name="{{ include.target }}-craftmanship" class="form-control" rows="10" id="craftmanship"></textarea>
```
robertfirek, line 60: Text area can also be `required`

_includes/apply_form.html
MiyamotoAkira: On this case we are not using required as we are really not submitting the form, but calling a js method that does an Ajax call.

_includes/apply_form.html
MiyamotoAkira: Same as above:
On this case we are not using required as we are really not submitting the form, but calling a js method that does an Ajax call.

PR #714, opened 2017-08-17
Update 2017-08-08-ddd-modeling.md

davidgrahamhall:
Updated 'Modeling' to 'Modelling'.

### DIFF
--- a/_posts/2017-08-08-ddd-modeling.md
+++ b/_posts/2017-08-08-ddd-modeling.md
@@ -1,8 +1,8 @@
 ---
 layout: post
 asset-type: post
-name: DDD Modeling from Alberto Brandolini
-title: 'DDD Modeling from Alberto Brandolini'
+name: DDD Modelling from Alberto Brandolini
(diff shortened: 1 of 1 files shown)


PaNaVTEC:
Why do we need double L? 

https://en.wikipedia.org/wiki/Domain-specific_modeling

davidgrahamhall:
Consistent with how he calls it: https://skillsmatter.com/courses/562-alberto-brandolini-s-ddd-modelling-workshop

PaNaVTEC:
In the official book it is with one L, this is an extract of the first page: 

```
The software development community widely acknowledges that domain modeling is central to
software design. Through domain modeling, software developers are able to express rich
functionality and translate that functionality into software implementation that truly serves the
needs of its users. Despite its obvious importance, however, there are few practical resources that
show how to incorporate effective domain modeling into the software development process.
```

davidgrahamhall:
Fair enough, lets stick with that! :-)

PR #718, opened 2017-08-21
Removed toby and resized Jorge's pic

mashooq:


PaNaVTEC:
Thanks for resize the image 👍

PR #732, opened 2017-09-06
making the default blockquote more prominent

mashooq:


assets/css/custom/custom-global-styles.scss
```diff
@@ -23,3 +23,15 @@ x:-moz-any-link {  /* 0 0 1 1 */
...
+
+blockquote {
+	border: none;
+	padding: 18px;
+	background: #f3f3f3;
+	border-left: solid 2px #666;
+}
+
+blockquote p {
+    font-size: 18px;
```
lpalma: I would prefer either `16px` or `14px` (default) for block quotes. The reason is that `18px` makes the quote stand out way too much, as if it wants to grab the reader's attention, which is not the purpose of quotes.

mashooq: I've changed the font size to 16. It's a compromise however, the purpose of the quotes is 2 fold. One as part of a narrative and explanation where I agree that it needs to carry similar weight to rest of the text. The other to provide a break from the monotonous text that our blogs currently have. It also serves purposes of a different kind of reader who's not interested is reading the blogs end to end is more pursing through various blogs.

lpalma: Interesting. I didn't consider the second perspective.

I agree that our blog (specially my latest post) is usually monotonous. Perhaps we can think of changing the layout, like giving more space between paragraphs, insert links to related blog posts, give more contrast between subtitles and paragraphs (this one I actually think is necessary), etc.

lpalma (approved):

assets/css/custom/custom-global-styles.scss
mashooq: Yes, certainly. Making the blockquotes moves us towards making longer texts more accessible. We can make these changes incrementally and even introduce some style guideline for blogs.

PR #733, opened 2017-09-07
Add liliana carneiro and resize ivan badia

nachogarcia:


lpalma:
This PR serves 2 purposes (as the title suggests).
We should keep our PRs as small as possible, doing one and only one thing. In this case, fixing the image size should have been a separate PR.

But no need to change it now :)

PR #738, opened 2017-09-15
Changed to gender neutral.

mashooq:


cohen990:
I'm aware that we have a lot of places that refer to "craftspeople" - but that language seems very clumsy to me - is "crafters" a reasonable alternative?

MiyamotoAkira:
I was going to talk about this with David and Matthew as part of the initiative circle. I do consider crafter/crafters the better option. Far less unwieldy.

mashooq:
In the last discussion at the catchup, we decided to use Craftsperson and Craftspeople. These are a matter of opinion I think both Craftsperson and Crafter are unwieldy replacements for Craftsman/Craftswoman. However, we need to choose one.

MiyamotoAkira:
Oh, I must have missed that. The need to choose a way was why I was going to have the meeting with David and Matthew (the other two on the circle) before the catch up this Tuesday about it. But if this has been talked about already ...

mashooq:
Happy for you guys to talk about it and see what fits better. I personally think both are clumsy so not too bothered which one we go with. Make sure you get Sandro involved in that discussion as well.

sandromancuso:
As @mashooq said, these changes are about what has been agreed by more than half of the company less than one month ago. If people still want to discuss this and change this decision we can change the website again. For now, I think this PR should be merged.

bnathyuw:
Hello!

First, thank you Mash for making this change. As you know, I think this is important, and I'm extremely pleased to see it.

To be honest, I don't have strong feelings about the choice of 'craftsperson' or 'crafter'. There are good arguments for each. Right now, I have a mild preference for 'craftsperson', which is convenient, as that seems to be the current decision.

I'm happy to give this PR my +1.

PR #748, opened 2017-10-02
Send recruiting emails to different accounts

nachogarcia:
Mergable after @mashooq changes the current Trello integrations


cyplo:
@mashooq, @nachogarcia - do we need anything more to merge this ? thanks !

nachogarcia:
I don't know if @mashooq had time to handle the Trello integrations. Let's wait for his confirmation.
But I agree, we should merge this sooner than later. Too many merges already :stuck_out_tongue_closed_eyes:

mashooq:
I'm looking into this now. I will need to create a new group join-us-bcn@codurance.com and have that forward to the BCN Trello Board.

nachogarcia:
Oh, I will then have to change the email address from 'join-us@codurance.es' to 'join-us-bcn@codurance.com'.
Let me know when it's done (just in case you decide to put other name) and I'll change it

PR #751, opened 2017-10-08
Added ability to search

MiyamotoAkira:
Basic ability to search on posts.

Probably we will want the search page to actually  be integrated inside publications, but is a starting point.


mashooq:
Hi Jorge, I tried this in the preview branch and it doesn't seem to work.

MiyamotoAkira:
Yeah, there is something happening on the deployed system. Works on my machine (ha ha). But for some reason it doesn't when deployed. Need to investigate further. I mentioned the issue on the slack channel

cohen990:
I wonder if this is related to #777 ?

MiyamotoAkira:
don't think so, the problem is related to the form action, which doesn't add the query parameter

cohen990:
I'm gonna close this unless anybody objects

PR #763, opened 2017-10-23
Add Richard Wild to website

richardjwild:


company/index.html
```diff
@@ -54,11 +54,13 @@
                   </p>
                   {{member.fullDescription}}
                   <ul class="list-unstyled list-inline team-socail">
+                    {% if member.twitter != null %}
```
cyplo: May be better to have these in a separate PR, what do you think ? To not mix content changes with the behaviour changes

cyplo:
Not sure about the photo - it's showing other people and the background is a bit distracting

company/index.html
richardjwild: Okay, I'll put this change in a separate PR and we'll need to merge that one first.

cyplo: cool, you can say so in the description and reference them both using #763 notation

richardjwild:
Fair comment on the photo. I've added a cleaner one and I think it fits with the other photos much better.

richardjwild:
Please no-one merge this PR until #764 is merged first.

PR #778, opened 2017-11-03
clean up local build system

cyplo:
Hello ! After the confusion around how to run the website locally reported by the people that touched the website recently, including me, I’ve created a PR to simplify the run/build process.
If you know how to do your local thing (on-machine, without Docker) - it should still work. The new recommended approach though is just to use docker and I’ve cleaned up the README to reflect that. The README had a lot of outdated info that was not true/did not work anymore - deleted it. 

The docker container start should be MUCH faster now as well, so should be possible to have everything running pretty quickly.


README.md
```diff
@@ -1,89 +1,15 @@
...
 
-To build and serve the site from Docker you need to build a local image based.
-Dockerfile for is ready to use in directory `docker`.
-
 You need to execute the following command to prepare an image and run a container:
 
 ```
-docker build --tag codurance-site:local -f ./docker/Dockerfile-local ./docker/
-docker run -d -v <<root_directory_of_this_project>>:/site -p 4000:4000 --name codurance-site codurance-site:local
+docker build --tag codurance-site:local -f ./Dockerfile-local . # you need to do this only once - this builds the container with all the dependencies for running the site
```
lpalma: What about moving the comments above the command? It would make it easier to see the separation

```
# you need to do this only once - this builds the container with all the dependencies for running the site
docker build --tag codurance-site:local -f ./Dockerfile-local .

# you need to run this every time you want to stand up the local server - it should watch for changes in the local files automatically though
docker run -i -t -v `pwd`:/site -p 4000:4000 codurance-site:local 
```

start_site.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+cd /site
+bundle install
+bundle exec rake servequick
```
lpalma: Also, we could pass a parameter to the second command (`docker run ...`) to pick the target we'd like to run (e.g. `servequick`, `en`, `serve`). Could be an ENV variable or something like that

lpalma (commented):
Worked like a charm and the README looks much better 👍 
Just a few minor comments, otherwise I'm happy to merge.

cyplo:
Fixed the things mentioned in @lpalma comments, thanks !

cohen990:
on my machine, when I edit files locally with the docker container running, the site does not get rebuilt

Attempting to reproduce

cyplo:
Feedback from @sandromancuso (in person) and the website slack channel: people would like to have the steps to run the website locally still in the readme but cleaned up and polished. After that we can merge this. Thank you for all the discussions !

cohen990:
merge before #784

PR #802, opened 2017-11-20
change description to reflect the post's description if available

cohen990:
using the tetris AI post as a lab rat


cohen990:
facebook was showing the old description because if I ask it to share one of our preview pages, it follow the canonical to our live site and picks up the description from there

PR #803, opened 2017-11-20
WIP: Use docker compose for local build

hoto:
* Simplified running docker by using docker compose:
    - Mounting only those files and folders which are necessary for the container.

* Removed the `--incremental` flag from `rake quickserve` to include all changes in the project (like adding a bio). This makes the site build in 4 sec instead of 2 sec, but we won't need to restart/rebuild the container every time.

* Removed unused stuff, like running tests (which we don't have).


.dockerignore
```diff
@@ -0,0 +1,6 @@
+.idea
+*.iml
+_site/
+_site_es/
+.bundle
+.jekyll-metadata
```
PaNaVTEC, line 7: Maybe we can add `.git` as well ?

docker-compose.override.yml
```diff
@@ -0,0 +1,41 @@
+---
+version: '3.1'
```
PaNaVTEC, line 2: Can you do the same with environment variables: https://docs.docker.com/compose/environment-variables/ and 2 small bash scripts `run-production-site.sh` and `run-development-shite.sh` you could achieve the same and it will be more readable.

hoto: You mean to replace yml files with shell scripts to make it more readable?

hoto: `docker-compose.yml` and `docker-compose.override.yml` are standard files recognized by docker compose.

`docker-compose.prod.yml` is a convention and is not really necessary here.

I could use just the `docker-compose.yml` and not create the other two yml files but it's a docker compose convention to have three separate ymls.

.dockerignore
hoto: added

cohen990:
Closing - low priority

PR #826, opened 2017-12-06
changed craftsmen to craftspeople

mashooq:


_includes/head.html
```diff
@@ -2,7 +2,7 @@
  <!-- Meta -->
 <meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
-<meta name="description" content="Our team of dedicated software craftsmen provides consultancy, software development, and training services to clients seeking high quality development processes and software solutions.">
+<meta name="description" content="Our team of dedicated software craftspeople provide consultancy, software development, and training services to clients seeking high quality development processes and software solutions.">
```
samanthalcohen, line 5: isn't it "provides" since the team provides? English is hard 😅

richardjwild: Ah god, I can see an argument both ways because a team is a single thing, which would make it "provides," but it refers to something that is plural (craftspeople) which would make it "provide".

It's difficult but I think the version in the pull request (provide) is correct.

samanthalcohen: if 2 agree then fair enough lol. I was reading the original and debating with myself about which one was correct. Just wanted to ask the question

richardjwild: I'll moige it.

PR #884, opened 2018-02-09
Fixing ALL-AUTHOR-POSTS button

rachelcarmena:
Fixing ALL-AUTHOR-POSTS button when author name is composed by more than 2 parts

### DIFF
--- a/_includes/publication.html
+++ b/_includes/publication.html
@@ -97,7 +97,7 @@
             <div class="media-body">
               <h4 class="h5 g-color-black g-font-weight-600">{{ page.author | author_links}}</h4>
               <p class="g-color-gray-dark-v5 mb-4">{{ member.fullDescription }}</p>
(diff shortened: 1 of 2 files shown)


lpalma:
I checked and indeed the button now links to the correct page :)

However, I think that the issue is not with the button itself, but rather with how the author's URL is being generated.

This is an example of the current url for Sergio's post:

`https://codurance.com/publications/author/sergio-rodrigo%20royo/`

and, actually, the browser displays the space character rather then the escaped version:

`https://codurance.com/publications/author/sergio-rodrigo royo/`

The button, on the other hand, is pointing to a URL with dashes instead of space:

`https://codurance.com/publications/author/sergio-rodrigo-royo`

Which, I believe, is how we should be generating our author's page.

Thoughts? :)

lpalma:
On a second thought, the button on production is broken and this PR solves the issue.

I'll raise another issue for correcting the Author's link

rachelcarmena:
Thanks @lpalma !! Please, take a look at the source code, because I understood what was happening when reading it: https://github.com/codurance/site/pull/884/files
I extracted a method, instead of a specific piece of code with this button.

PR #910, opened 2018-03-16
blog on interfaces with single implementations

mashooq:


rachelcarmena (approved):
I loved it! I lived a similar situation a few days ago :wink:

PR #912, opened 2018-03-20
Blog post: Getters and Setters Considered Harmful

richardjwild:


rachelcarmena:
Wow @richardjwild ! I enjoyed a lot with your post. Congrats! My favorite "quote" from your post: _Practices are not set in stone, they evolve over time and bad practices can be fixed._

PR #929, opened 2018-04-12
Update Dockerfile with Jekyll-pug

### LINKED ISSUE #928: Can't build from docker
I just tried to build the site following the docker instructions. There was some issue with a `.pug` file. When I ran it with native ruby, it told me to run `npm install --global pug`

Possibly a quick fix in the dockerfile?

lpalma:
Fixes #928

### DIFF
--- a/Dockerfile-local
+++ b/Dockerfile-local
@@ -3,6 +3,8 @@ FROM ruby:2.3.1
 RUN apt-get update 
 RUN apt-get install -y --no-install-recommends apt-utils
 RUN apt-get install -y locales
+RUN curl -sL https://deb.nodesource.com/setup_9.x | bash -
+RUN apt-get install -y nodejs
 RUN apt-get clean
 RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen
 RUN dpkg-reconfigure --frontend=noninteractive locales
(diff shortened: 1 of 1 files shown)


Dockerfile-local
```diff
@@ -14,6 +16,7 @@ RUN gem install jekyll
 ADD Gemfile /tmp/Gemfile 
 ADD Gemfile.lock /tmp/Gemfile.lock
 
+RUN npm install pug -g
```
hoto: I would put this before any `ADD` so docker have a chance to cache this layer. 
Could be right after `RUN apt-get install -y nodejs`

Dockerfile-local
```diff
@@ -3,6 +3,8 @@ FROM ruby:2.3.1
 RUN apt-get update 
 RUN apt-get install -y --no-install-recommends apt-utils
 RUN apt-get install -y locales
+RUN curl -sL https://deb.nodesource.com/setup_9.x | bash -
+RUN apt-get install -y nodejs
 RUN apt-get clean
```
hoto: This line is funny, because I think this actually makes the image larger instead of smaller.
But let's not save the world when it does not need to be saved.

hoto (approved):

Dockerfile-local
samanthalcohen: how could it make the image larger?

PR #931, opened 2018-04-13
add marco description and photo to site

tonycodurance:


keith-smale:
I think there is a problem with the colon in the last paragraph tag of your description. Jekyll doesn't seem to like it.

keith-smale:
Yep definitely a problem with the colon. Also the image needs to be uploaded as 400px by 400px so it matches the others in the grid.

keith-smale:
Photo re-sized and colon removed from description which was blocking the build.

PR #946, opened 2018-05-25
Cookie policy

keith-smale:
Sites who use Cookies should be compliant with eu cookie law which came into effect in 2011. Whilst our site uses Cookies we have no banner advising users that we use cookies nor do we have a cookie policy explaining the cookies the site uses and their expiry timeframes, there is only a mention of cookies in our privacy policy.

I've added a cookie policy to the site and used the following repo, https://github.com/studio24/cookie-message, to add some JS to support the cookie banner.

Issue - I've noticed that the once the banner has been shown and the cookie relating to the display of the cookie message has been placed on the device the banner flashes up when new pages are loaded, i've placed the scripts as the repo advises to avoid this but it's still happening, so not sure what the problem is.

If we choose to merge this, I would like someone (David) to check the cookie policy wording and the cookie information is correct before it's added to the site.


_i18n/es.yml
```diff
@@ -88,6 +88,7 @@ footer-useful_links:
    link_4_title: Carreras
    link_5_title: Acerca de
    link_6_title: Política de privacidad
+   link_7_title: Política de Cookie
```
PaNaVTEC: Política de Cookie*s*

keith-smale: updated :)
Thanks

Mattsi-Jansky:
What's the state of this PR @keith-smale? I can review if that'd help

keith-smale:
Hey, not sure to be honest. 
I added this cookie-banner a while back as I couldn't see we had something like that but were using cookies, so probably not compliant with the EU Cookie Law. 

I raised the PR for someone to check what I had added to make sure it was ok and make any adjustments. I got it from another repo so may needs some tweaks. 

Seems to be a little flicker of banner once it has been acknowledged the first time.
Also it may be a little out of date.

alfredodev:
I merge the changes and I've fixed the flicker.

I think that we need add to the cookie policy page information about sharethis, vimeo and youtube at least, but I can create a new issue for this.

If you see it alright @keith-smale @davidgrahamhall I think this is production ready

Mattsi-Jansky:
Tested on Chrome and Firefox Nightly, flicker does seem to be solved. Good job @alfredodev. I did notice that it automatically adds the cookie to accept the banner though- should we have an accept button to confirm the user's consent? Seems to be the standard elsewhere, though I don't know what the legal requirements are exactly.

keith-smale:
@Mattsi-Jansky I don't know whether we should have the button.

Since GDPR I've also seen people giving the user the control to turn off some cookies, but as we have nothing this solution should still probably be ok.

Would want @davidgrahamhall or @steve-codurance @stevelydford to be happy with these changes , cookie policy wording and privacy policy wording before this is merged.

alfredodev:
I've implemented this for various Microsoft pages and they have this message
"This site uses cookies for analytics, personalized content and ads. By continuing to browse this site, you agree to this use."
And the message is removed once you click any link.
We can do this.

keith-smale:
Cool so if we are all happy. Then just need to get some of David's or Steve's time to check it all over and merge it.

alfredodev:
I remembered now that we also didn't add cookies until consented, which is not that easy...
I'm reading a bit and everyone says different things.
Let me investigate it further.

alfredodev:
Also I've got this report
[report-codurancecom-1246176.pdf](https://github.com/codurance/site/files/2400879/report-codurancecom-1246176.pdf)

alfredodev:
So everybody says something different.
For example, [here](https://www.itgovernance.eu/blog/en/how-the-gdpr-affects-cookie-policies) they say that implicit consent is not valid anymore:

> Implied consent is no longer sufficient. Consent must be given through a clear affirmative action, such as clicking an opt-in box or choosing settings or preferences on a settings menu. Simply visiting a site doesn’t count as consent.

[Here](https://www.analyticsmania.com/post/gdpr-cookie-consent-notification-with-google-tag-manager/) they say, that accepting it by navigation is valid as a "soft opt-in"

> Here’s the flow:
> 
> A new visitor lands on a page and sees the cookie notification banner at the bottom of the screen.
> Banner offers to accept cookies or decline them by going into cookie settings. Until this moment no tracking tags are fired yet. In other words, a visitor still has the opportunity to act before cookies are stored in his/her browser.
> If a visitor clicks Accept Cookies button, this is a consent.
> If a visitor continues browsing a website, this is considered as a soft opt-in and all the cookies are also stored in the browser.
> If a visitor clicks on Cookie settings he/she can change preferences and opt-out of particular groups of cookies.

And big companies like Microsoft do this.
So I think we should do that.
I changed the message to add that by navigating you are giving consent.

With that we can merge and close this, but to be compliant we need to do this too:
#1063
#1064 
#1065 
#1066 
#1067 
#1068

keith-smale:
This the Information Commissioners Office guidance [here](https://ico.org.uk/for-organisations/guide-to-pecr/cookies-and-similar-technologies/). 
These guys set and apply the rules so this is what should be followed. 

In relation to consent they say:
> To be valid, consent must be freely given, specific and informed. It must involve some form of unambiguous positive action – for example, ticking a box or clicking a link – and the person must fully understand that they are giving you consent. You cannot show consent if you only provide information about cookies as part of a privacy policy that is hard to find, difficult to understand, or rarely read.
>
> Consent does not necessarily have to be explicit consent. However, consent must be given by a clear positive action. You need to be confident that your users fully understand that their actions will result in specific cookies being set, and have taken a clear and deliberate action to give consent. This must be more than simply continuing to use the website. To ensure that consent is freely given, users should be able to disable cookies, and you should make this easy to do.

So we may need to do some more by adding a button and not actually dropping the cookies until the consent is given.

alfredodev:
Ok, should we change the style then and make it like bigger and fixed so there is more people accepting it?
Something like those options here:
https://cookieconsent.insites.com/app/themes/insites-cookie-consent/examples/example-1-themes.html

alfredodev:
New sticky banner with accept button added

MiyamotoAkira:
I'm going to merge this one. If any fixes are needed afterwards, then we will deal with them then.

PR #962, opened 2018-06-18
Minor layout and spelling fixes

Mattsi-Jansky:
Mostly raised to fix the following layout issue: https://imgur.com/SrttBOY 
^ Only repros in Firefox. Some "how we work" elements overflowed and had no max height, so
they weren't all consistent in size.

Also fixed a couple other very minor things while I was there, markup indentation and a typo.

I'm not sure which conventions for naming etc we're folllowing in this project, or whether I've placed the style in the right file. Let me know if it doesn't follow convention and I'll change it.

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -215,7 +215,7 @@ training-software-architecture-and-cloud:
    link_title: See our courses ...
 training-technologies:
    title: Technologies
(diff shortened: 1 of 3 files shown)


Mattsi-Jansky:
Correction: It also repros in Chrome, but only at lower screen widths than in Firefox.

PR #992, opened 2018-07-31
Cell pattern

cohen990:


rachelcarmena:
I loved the analogy, the idea for organizing the source code and the idea for progressing the pattern. Congrats @cohen990

PR #1005, opened 2018-08-12
Use of automated coding style tooling

MiyamotoAkira:


_posts/2018-08-13-automated-code-styling.md
```diff
@@ -0,0 +1,62 @@
...
+date: 2018-08-11 08:01:00
+description: Where I describe my issues with tools like Stylecop and Sonarqube for documentation comments.
+image: 
+    src: /assets/custom/img/blog/2018-08-13-automated-code-styling/caffeine-coffe-cup-6347.jpg
+tags: 
+- software processes
+---
+# Intro
+
+Recently I have had the "privilege" of going through some Javadoc on a project that needs to move from Java 7. The linter for javadoc was tightened for Java 8. Of course, the project did fail to build on Java 8. It fail on so many javadocs.
```
Mattsi-Jansky: Grammar: "Of course, the project did fail to build on Java 8. It fail on so many javadocs." -> "Of course, the project failed to build in Java 8. It failed on many javadoc entries."

_posts/2018-08-13-automated-code-styling.md
```diff
@@ -0,0 +1,62 @@
...
+/// <summary>
+```
+
+Both are enough to pass SonarQube or Stylecop (rules can be tightened a bit). But I hope we can agree that they are not useful and a waste of screen space and developer effort. Furthermore, there is a lack of subtletly and handling of special cases on tools that do check coding style. Because sometimes you need to apply logic that is just too specific for one example to be worthy of spending time codifying that "rule".
+
+# The Now
+
+I learned to dislike using an automated tool for documentation comments. They are just not fit for the task. I have my own thinking about when to use documentation comments:
+
+- First, only use documentation comments on a public API, when people are going to use your library without being able to easily see the source code. I will not consider having the repo open to qualify as easily. Only if the code is in the same project can be considered so.
```
Mattsi-Jansky: The last two sentences are a bit awkward to parse, not sure what the correct grammar is but perhaps: "I don't think a repo being open is enough to qualify as easily accessible; only if the code is in the same project is it easily accessible."

_posts/2018-08-13-automated-code-styling.md
MiyamotoAkira: Is not the first one correct when trying to emphasize the fact that it happened? Maybe I'm getting confused there.

_posts/2018-08-13-automated-code-styling.md
MiyamotoAkira: Redone slightly. Can you check if the new version reads better?

_posts/2018-08-13-automated-code-styling.md
Mattsi-Jansky: The first sentence may not be technically incorrect, not sure myself, but it's very uncommon to see "did fail" instead of "failed" and may break the reader's immersion.

_posts/2018-08-13-automated-code-styling.md
Mattsi-Jansky: Yep, reads much better :+1:

PR #1013, opened 2018-08-17
The Functional Style part 2 - iteration 1

richardjwild:


MiyamotoAkira:
Could you add an indicator for the language that you are using? If the whole series was on a single one, that would be fine. But jumping between languages, adding a note just before the code would make it much easier to switch context or even to know which context you are on

MiyamotoAkira:
Also, let me point to the Functional Calisthenics and the rule of "No explicit recursion". Do you use map/reduce later on the series? Because I can see an awesome refactoring of the examples to use reduction :innocent:

rachelcarmena:
Another great post @richardjwild ! I agree with @MiyamotoAkira about an indicator about the language in every piece of source code.

In case you would like to include a tail recursive version about factorial: 

```
private long fact(long sum, long n) {
    if (n == 0)
        return sum;
    else
        fact(sum * n, n - 1);
}

private long factorial(long n) {
    fact(1, n);
}
```

And in case you would like to include a reference to another post by @PaNaVTEC : https://codurance.com/2017/12/07/recursion/

And congratulations for this series of posts again! Looking forward the following one!!

Signed: a recursion lover :wink:

rachelcarmena (approved):

richardjwild:
@MiyamotoAkira 

> Also, let me point to the Functional Calisthenics and the rule of "No explicit recursion". Do you use map/reduce later on the series?

Yes, that is coming in the next two articles. There will be no more recursion in the rest of the series. I don't consider recursion part of the "functional style". I will bear that in mind and try to emphasise that in the coming articles.

PR #1017, opened 2018-08-28
Crafting Webapps training

nachogarcia:


rachelcarmena (commented):
Awesome @nachogarcia!! 
I think there are a lot of contents for 2 days. Maybe we can propose more days and the client can decide the contents. For example, maybe they don't need SEO or refactoring.
And I only found confusing "the most important concepts for a webapp". I'd prefer "for developing a webapp", because we don't include usability or UX (more related with design), for example, and they are very important as well.

nachogarcia:
Those points are the ones that would ideally be covered if everything went smooth and the atendees had a really good base.
Probably a couple of those points will just be discussed (as in how does the SEO work with SPA or how the atendees could try to attack their monolith), and going deep if they show interest.
I don't know how the points of the other courses correlate with the deepness of the content, so I might be in the wrong by adding them of course, it's open for discussion.
Also I don't like the idea of opening the course for variable days; it's flexible enough by giving choice of libraries. Maybe in the future, but at the moment it would be too much planification for me.
I like keeping in mind every part involved on a product, so I'll definetly look at the UX part.

Thanks a lot @rachelcarmena!

PR #1040, opened 2018-09-10
Fix language links

### LINKED ISSUE #1039: Language (ES/EN) links don't work
This is something Mash asked me to have a look at, and as far as I can see there isn't an issue for it yet (#548 seems to be about a separate issue).

When I: Click the ES link from the .com site or the EN link from the .es site

I expect: To follow a link to the relevant domain (.com -> .es, .es -> .com)

What actually happens: Nothing. (`href="/"`)

Mattsi-Jansky:
Resolves #1039

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -17,8 +17,6 @@ GEM
     commonmarker (0.17.7.1)
       ruby-enum (~> 0.5)
     concurrent-ruby (1.0.5)
-    diff-lcs (1.3)
-    diffy (3.2.0)
     ethon (0.11.0)
       ffi (>= 1.3.0)
     execjs (2.7.0)
@@ -251,19 +249,6 @@ GEM
     ref (2.0.0)
     regexp-match-polyfill (1.0.2)
     rouge (2.2.1)
-    rspec (3.7.0)
-      rspec-core (~> 3.7.0)
-      rspec-expectations (~> 3.7.0)
-      rspec-mocks (~> 3.7.0)
-    rspec-core (3.7.1)
-      rspec-support (~> 3.7.0)
-    rspec-expectations (3.7.0)
-      diff-lcs (>= 1.2.0, < 2.0)
-      rspec-support (~> 3.7.0)
-    rspec-mocks (3.7.0)
-      diff-lcs (>= 1.2.0, < 2.0)
-      rspec-support (~> 3.7.0)
-    rspec-support (3.7.1)
     ruby-enum (0.7.1)
       i18n
     rubyzip (1.2.1)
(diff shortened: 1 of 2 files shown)


MiyamotoAkira:
Coolio

MiyamotoAkira:
Be aware that if you use docker, you want to redo the images, as the images had the gem already applied to them. Although on this case there is no need to do it right now

MiyamotoAkira (approved):

alfredodev:
I think this should have the base url:

                  <li> <a class="d-block g-color-main g-color-primary--hover g-text-underline--none--hover g-py-8 g-px-20" href="{{ domain[1] }}{{ page.url }}">{{ domain[0] | upcase }}</a> </li>

VS

                  <li> <a class="d-block g-color-main g-color-primary--hover g-text-underline--none--hover g-py-8 g-px-20" href="{{ domain[1] }}{{ baseUrl }}{{ page.url }}">{{ domain[0] | upcase }}</a> </li>

Mattsi-Jansky:
@alfredodev `domain[1]` already contains the baseUrl. It's value can be either `https://codurance.es` or `https://codurance.com`

alfredodev:
What about https://codurance.es/folder as in for the PR deplyments we were looking at before

Mattsi-Jansky:
Hum, interesting point. The thing is that it's linking into a different instance of the website, so our baseurl doesn't tell us anything about the website it's going to. For example, `http://codurance-site-pr.s3-website-eu-west-1.amazonaws.com/site-fix-language-links/publications/` has a folder in the path and is in the publications sub-domain. If you click through to the ES domain, it'll redirect correctly and put you in `https://codurance.es/publications/`. If the baseUrl were there it would wrongly redirect to `https://codurance.es/site-fix-language-links/publications/`.

However, there may be instances where the site we're linking to is in a folder. I think the right thing to do in that case would be to modify the domain strings in `_config.yml`, eg change `es: "https://codurance.es"` to `es: "https://codurance.com/es/"` or something.

alfredodev:
I think this would happen because we are not deploying the ES version of the site for the PR, but we should, as we need to review the ES version too before deploying.
If we do this, the change should be from:
https://codurance.com/site-fix-language-links/publications/
to 
https://codurance.es/site-fix-language-links/publications/

I think all the absolute url's should have the baseUrl, we have already some broken images on the PR deploys, but we can add another issue to review all of this together.

Mattsi-Jansky:
We should definitely deploy both, and it'd be good to have their language links point to one another. We would need to find a way to parameterise the `domains` variable on deployment but could be done.

Issue #548, opened 2017-01-25
Link to Spanish version of the site always points to the absolute link


neomatrix369:
Every page on the site has an English [EN] and a Spanish [ES] version and we can switch between them by clicking on the **[EN]** or **[ES]** links.

But the **[ES]** link always points to the `http://xxxx.es` site even when the site is built on the local machine or via the preview site (temporary S3 bucket).

This is incorrect and it prevents viewing of the Spanish version of the site hosted on the local machine or the preview site _via these links (ES)_. Although we can still see the Spanish pages by added a `/es` just hostname of the local machine or the preview site URL.

PaNaVTEC:
Since we have now base urls, one possible solution could be to deploy both pages to the same bucket but in different folders. Something similar to:

```
english site: bucket root # codurance.com like it is now
spanish site: bucket root/es # codurance.es redirecting to codurance.com/es or bucket root/es 
```

If we have that configuration we can have relative links in the top right language menu

PR #1057, opened 2018-09-20
Remove python libs dependencies by hardcoding domains temporarily

alfredodev:
Hardcoded domains temporarily to fix the build.
Also I needed to review that already for #548


MiyamotoAkira:
Merging with the comment that a ruby library should be used.

alfredodev:
Created #1058

PR #1078, opened 2018-09-29
Safety musings

MiyamotoAkira:


MiyamotoAkira:
DO NOT MERGE. I will once it has been reviewed

_posts/2018-10-01-safety-musings.md
```diff
@@ -0,0 +1,75 @@
...
+---
+
+So here I am, again, this time with some serious musing around what we do as engineers/developers/craftpeople. Our languages, tools, and techniques, determine what we can do, how good our systems are, how safe, how secure, and how fit for the task. Because this affects people's lives we need to be careful on what we do. Robert Martin (Unble Bob) has been talking about the necessity of a code similar to the Hippocratic oath. I am of the believe that it will come the time in which we will need to be part of a professional body to perform our duties (after all, the Hyppocratic oath did not stop the need for the Medical license)
+
+## Discussing about what are functional languages
+
+The other night I was discussing with some colleagues what is a Functional Language. On their definition, only pure statically typed languages can be called Functional Languages. No IO Monad, no functional language. My contention was that immutability (and with it the reduction of side effects) is enough for a language to be called functional. Their thinking eliminates Clojure, Elixir, Erlang and a few others. Instead of seeing functional languages as a spectrum, it was an all or nothing.
```
CarlosMChica: I assume that you are referring to our chat the other day. I would like to clarify something that maybe I didn't communicate correctly. 

I don't necessarily meant that you need an IO monad to do functional programming, but a way of controlling side effects with the type system. IO monad is a way of doing it and probably the most popular, but not the only one

richardjwild:
I liked it, it is still in a rough state in patches but the overall ideas I agree with. The safety you are talking out comes from constraints. The purpose of static typing is to constrain us, and that is useful because it eliminates whole classes of programming errors. We are also constrained by the three programming paradigms, none of which really add features but rather they _remove_ freedoms:

1. Structured programming imposes constraints on direct transfer of control (GOTOs). It taught us that all algorithms can be broken down into sequence, selection, and iteration, and they can be implemented with control structures that are easier to comprehend than tangled spaghetti code (there's the safety feature).
2. Object-oriented programming imposes constraints on indirect transfer of control, i.e. dynamic dispatch (function pointers). Function pointers are useful for creating abstractions and inverting dependencies. So OO programming provides a means for achieving these things, managed by the compiler, while removing the ability to create pointers to functions so that you can't shoot yourself in the foot (safety feature).
3. Functional programming imposes constraints on mutating state. Instead of changing the values of things unrestrained, the language imposes discipline so that you know where the mutating state lives, and the language additionally protects you against synchronization problems (safety feature).

Also, when the three paradigms are understood in this way - as safety features that _remove_ unrestricted features and replace them with _managed_ features - then this supports your opinion of what the essence of functional programming is. Because the alternative definition proposed by our colleagues, strong typing and IO monads, doesn't work so well. Static typing is indeed a safety feature but does not fit particularly as a definition of FP because you can have strong static typing and still mutate state all over the place. These two concerns are orthogonal.

_posts/2018-10-01-safety-musings.md
richardjwild: Clojure controls side effects but not through the type system, it does it simply by lacking an assignment operator.

Personally I think the strong vs. dynamic typing spectrum is totally orthogonal to the imperative vs. functional spectrum.

MiyamotoAkira: It was indeed all steaming from our conversation. I will include your addendum. But to be honest, the thing is that because of that way of thinking I started to ponder about it, and in parts actually agree with you, and from that I evolved the whole post :-)

MiyamotoAkira:
Interestingly  I remember a chat from Uncle Bob in which said that first came Functional, then OOP then structured (in only a few years). But the adoption was on the reverse. We need to be able to do structured programming before we were able to do Object orientation. And we needed to do that before we were able to do FP (we are talking about the general programmer here, not the super duper genius ones).

If you don't mind @richardjwild , I am going to add your ideas plus the above in some form to the post.

MiyamotoAkira:
By the way, yes, although I wasn't clear about it, I do believe typing and paradigm are separate independent dimensions. Will improve that as well.

MiyamotoAkira:
Thinking that the section about techniques is thread bare.

_posts/2018-10-01-safety-musings.md
CarlosMChica: I do agree with this sentence:

"Personally I think the strong vs. dynamic typing spectrum is totally orthogonal to the imperative vs. functional spectrum."

However I also believe the following is true: 

"One doesn't really have control of side effects unless one is constraint from calling inpure code from pure code. Once you invoke an effectful function from a pure function, any callee of it and its ancestors become inpure." 

Honestly I don't know how you can do that without types, but if possible then I would consider that as valid as a typed solution. 

The sentence above have a lot of implications in how you write and arquitect FP apps in the large. It also summarizes how I believe FP code should be written and even though this is very opinionated, any code that doesn't follow that I wouldn't consider fully functional.

We can argue about being pragmatic about the rule, which probably applies here in a way, but still then we would be losing the ability to reason about effects in a safe manner in that part of the code. Whether you need/want to do that in a particular application can also be argued.

My view goes more in the direction of being able to control side effects and needing types to do that, than considering static/dynamic code functional/not functional.

BTW I wish we had more chats like this more often, I enjoy this :)

CarlosMChica: Maybe an example would help.

Taking the above to the exteme, printLn has side effects, therefore any code that calls printLn and all its ancestors now have side effects and are therefore not pure.

MiyamotoAkira: PrintLn is the less interesting of the side effects :-) Something like a web call or a database query are more interesting to show as an example.

I have expanded the idea of types, as they are four kind of types available to languages (check the updated blog).

PR #1088, opened 2018-10-05
Delete Matthew from site.

bnathyuw:
’Tis my last day, so it’s time to apply the Campsite Rule and tidy up after myself.


chriseyre2000 (approved):
Looks good.

PR #1093, opened 2018-10-08
Attempting to speed deploy by splitting assets into own job

chriseyre2000:
Split the assets into it's own step that can be manually triggered (or not).


MiyamotoAkira (approved):
All cool

PR #1103, opened 2018-10-10
Added some documentation for using backstopjs.

chriseyre2000:
Added some starter documentation for backstopjs.

### DIFF
--- a/tools/snapshots/README.md
+++ b/tools/snapshots/README.md
@@ -1,0 +1,32 @@
+# Snapshot Testing: backstopjs
+
+This is a tool that can be used for regression testing of css changes.
+
+The tool that we are using is backstop.js
+
+```
+npm install -g backstopjs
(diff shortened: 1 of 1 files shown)


MiyamotoAkira:
I use the docker instance to run the app. Could this be run from docker?

chriseyre2000:
It does have docker options:

https://github.com/garris/BackstopJS

Mattsi-Jansky:
Do we need to add `npm install -g backstopjs` to `Dockerfile`?

MiyamotoAkira:
Ideally yes. What we want is to avoid unnecessary installations on every single change we do in a post.

chriseyre2000:
I have created an issue to create the docker image for this: #1131

This is more useful while refactoring css rather than on a build server.

This PR adds some documentation.

PR #1111, opened 2018-10-10
Moved Bartomeu picture that was missplaced after the big merge

alfredodev:


Mattsi-Jansky:
Woops. LGTM, so long as you investigate the CI failure.

PR #1115, opened 2018-10-10
Use bigger logo

alfredodev:
Before
image
After
image


Mattsi-Jansky:
Much better 😍Is there an SVG version we could possibly use?

PR #1143, opened 2018-10-17
Expand tests to image helper

chriseyre2000:
We now have unit tests for the image helper.

These changes included adding a validation rule that will fail the build if the alt text is empty or missing.

### DIFF
--- a/spec/_plugins/image_helper_spec.rb
+++ b/spec/_plugins/image_helper_spec.rb
@@ -2,6 +2,27 @@ require 'jekyll'
 require 'jekyll/tagging'
 require_relative '../../src/_plugins/image_helper.rb'
 
+class DummyImageConfig
+  attr_reader :config
+
+  def registers
+    {:site => self}
+  end  
+
+  def config
+    { 
+      'baseurl' => 'baseurl',
+    }
+  end  
+end
+
+module Jekyll
+    class ImageTag < Liquid::Tag
+       attr_accessor :context    
+    end    
+end    
+  
+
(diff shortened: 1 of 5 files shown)


chriseyre2000:
Currently these images do not work on the preview site.
I am going to fix this.

This is the problem article: 2017-06-04-ncraft-in-20-pictures.md

spec/_plugins/image_helper_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    it 'Can render a simple image with alt text' do
+      token = double
+      expect(token).to receive :line_number
+
+      tag = Jekyll::ImageTag.send :new, 'img', '/assets/custom/img/blog/ncraft-2017/event-storming-with-brandolini.jpg "ALT"', token
+
+      expect(tag.render(nil)).to eq('<p></p><img src="/assets/custom/img/blog/ncraft-2017/event-storming-with-brandolini.jpg"  alt="ALT" title="ALT" class="img img-center img-fluid style-screengrab">')
+    end
+    
+    it 'will throw an exception when an img does not have a non-blank alt tag' do
```
alfredodev: double negation here "does not have a non-blank alt"
What do you think about this? 
"will throw an exception when an img has a blank alt tag"

alfredodev:
Nice alt messages!

spec/_plugins/image_helper_spec.rb
chriseyre2000: Renamed.

PR #1150, opened 2018-10-23
Clean backstop

chriseyre2000:
This gives us a baseline of backstop tests that pass.

I have had to ignore a lot of differences, but dancing animations do that.


chriseyre2000:
This was too much of a change. Abandoning.

PR #1153, opened 2018-10-26
Can I zip everything?

chriseyre2000:
This should zip all assets which should make a huge difference to the site.


chriseyre2000:
This is too simplistic to work. We would need to configure the cdn to properly set the headers this would break all the other preview branches. AWS documentation is also typically slightly broken.

PR #1155, opened 2018-10-31
Improve ruby tests

chriseyre2000:
We now have 100% Ruby test coverage.


spec/_plugins/author_generator_spec.rb
```diff
@@ -26,12 +26,110 @@ def config
...
+
+  it 'fails to generate if no author index' do
+    site = double 
+    allow(site).to receive(:config).and_return({})
+    allow(site).to receive(:layouts).and_return({'no_author_index' => nil})
+    genAuthor = Jekyll::GenerateAuthor.new   
+    expect{genAuthor.generate(site)}.to raise_exception(UncaughtThrowError, /uncaught throw "No 'author_index' layout found."/)
+  end  
+
+  it 'GenerateAuthor calls write_author_indexes with no posts and no collections' do
```
alfredodev: I can't see an expectation of a call  to write_author_indexes as the name suggests, but it's probably my lack of knowledge of ruby

spec/_plugins/author_generator_spec.rb
```diff
@@ -26,12 +26,110 @@ def config
...
+    
+    author_index_writer = double
+
+    allow(Jekyll::AuthorIndexWriter).to receive(:new).and_return(author_index_writer)
+
+    genAuthor = Jekyll::GenerateAuthor.new       
+    genAuthor.generate(site)
+  end
+
+  it 'GenerateAuthor calls write author with one author from each' do
```
alfredodev: Maybe use the same format here or on the previous test
> calls write_author_indexes
VS
> calls write author

alfredodev:
I've added a couple of comments on really small things, other than that is ok, feel free to merge it once the image is updated.

Issue #1156, opened 2018-11-01
Outdated SC London 2018 banner


CarlosMChica:
I was reading a publication on our website yesterday and was hit with a banner about SC London 2018 which has already happened, we should probably remove this banner or change it for something more useful.

![screenshot_2018-11-01-13_49_24](https://user-images.githubusercontent.com/3494156/47855675-2f8bc480-dddd-11e8-963a-eb192dd7315f.png)

Mattsi-Jansky:
Perhaps we should think of a way to avoid this happening in the future, too. Is there some way we could set up a notification/reminder to remove it? Wrapping it in a clause to only display until a certain date would be good too, but we'd still need to remember to come back and clean it up.

alfredodev:
This is a starting point #1158 
We need to talk to @davidgrahamhall to see what more to do with this.

PR #1158, opened 2018-11-02
Hide SC lodon banner

alfredodev:
Related to #1156
I've decided to maintain the template for the next banner and hide it depending on a date, in this way we can maintain the condition and only change the date for the next banner.


alfredodev:
You can see the banner not appearing here:
http://codurance-site-pr.s3-website-eu-west-1.amazonaws.com/site-remove-sclondon-banner/2015/07/18/cqrs-and-event-sourcing-for-dummies/
versus:
https://codurance.com/2015/07/18/cqrs-and-event-sourcing-for-dummies/

PR #1172, opened 2018-11-28
Change Scott's role and description

scotteza:


upsd (changes requested):
There is a slight typo: `enjoys working C# and SQL` -> `enjoys working with C# and SQL`.

upsd (approved):

PR #1180, opened 2018-12-09
Fix: removing images with text and unnecessary alternative texts

rachelcarmena:
Hi colleagues, I made several mistakes when writing my posts. I hope this is the last one in order not to disturb you. 
In this case, I'm removing images with text and unnecessary alternative texts (an empty alt attribute is valid when it's not necessary to provide the information of the existence of that image). 
Thanks in advance!

### DIFF
--- a/src/_posts/2017-09-01-do-we-have-a-good-safety-net-to-change-this-legacy-code.md
+++ b/src/_posts/2017-09-01-do-we-have-a-good-safety-net-to-change-this-legacy-code.md
@@ -36,14 +36,14 @@ So, we should aim at killing every mutation with tests.
 When I heard about it I thought about that game I played when I was just a teenager: _Super Pang_.
 
 <center>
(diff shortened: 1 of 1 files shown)


MiyamotoAkira (approved):

MiyamotoAkira (changes requested):
Just having a second look. Maybe is better to have the actual lines inlined within the posts, using triple quote, rather than pointing to the gist.

rachelcarmena:
Good idea, @MiyamotoAkira ! I gonna change it!

rachelcarmena:
Done @MiyamotoAkira ! Merging with remote repository (the additional one), rebasing, amending and pushing with force. So, PR with only one commit again. I know you like a clean commit history :wink:

MiyamotoAkira (approved):

rachelcarmena:
Thanks @MiyamotoAkira !! :dancer:

PR #1196, opened 2019-01-11
Change volumes paths in docker-compose.override file

ghost:
This fixes the watch in the docker container so you don't have to build every time you make a change.


MiyamotoAkira (approved):
Coolio!!!!

PR #1236, opened 2019-02-20
build and upload preview in both spanish and english

jaramir:


build/config/_config_prb.yml
```diff
@@ -1 +1,3 @@
-baseurl: /something
+baseurl: /site-branch-name/es
```
alfredodev: That this has only the Spanish settings confused me.
I went to look at it and seem's like this is used more like as an example?
Can we remove it? use example settings?

jaramir: oops. I've returned it to its original state

PR #1277, opened 2019-04-02
Rewording to emphatise delivery and reduce the use of 'helping' word

WencesLlobet:
Also:
Add M&S description.
Removed unreferenced descriptions that where not published.
Reword trinity mirror description to reflect that our work there is over.


hoto:
Is this still valid? Seems to be forgotten.

PR #1325, opened 2019-05-30
Make the intro in blog "microservices adventures" more formal

hoto:


hoto:
I'm leaving the old version.

PR #1329, opened 2019-06-07
add author name, twitter, and email to atom.xml feed

jaramir:


src/atom.xml
```diff
@@ -13,17 +13,31 @@ title: Atom Feed
         <email>{{ site.email }}</email>
     </author>
     <rights>Copyright (c) {{ site.sitename }}</rights>
-    {% for post in site.posts limit:15 %}
+    {%- for post in site.posts limit:15 -%}
```
MiyamotoAkira, line 16: Why the hyphens?

jaramir: suppresses the newlines https://shopify.github.io/liquid/basics/whitespace/

MiyamotoAkira: Oooh

PR #1358, opened 2019-06-27
Insert code in header to install Google Tag Manager

bourkemeister:
Inserted the standard text that Google Tag Manager asks you to insert into the header for each page.

### DIFF
--- a/src/_includes/head.html
+++ b/src/_includes/head.html
@@ -1,4 +1,4 @@
-<title>{% if page.title %}{{ page.title }}{% else %}Software Craftsmanship and Agile Development{% endif %} | Codurance | Craft at Heart | London | Barcelona</title>
+    <title>{% if page.title %}{{ page.title }}{% else %}Software Craftsmanship and Agile Development{% endif %} | Codurance | Craft at Heart | London | Barcelona</title>
  <!-- Meta -->
 <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


alfredodev:
We need to check if we need to do something special about GDPR with this.
For example, with Google analytics we anonymise the ip address until the user gives consent.

src/_includes/head.html
```diff
@@ -111,6 +111,13 @@
     ga('create', '{% t analytics.google-id %}', 'codurance.com');
     ga('set', 'forceSSL', true);
 
+    // code for installing Google Tag Manager
+    (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+    'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+    })(window,document,'script','dataLayer','GTM-WMKBC5G');
```
alfredodev, line 119: We should put this in the config file as the google-id for analytics
`analytics.google-id`

LL782:
Closing this as it's gone stale

PR #1361, opened 2019-06-28
Update readme with a workaround for mounted file permission problem

hoto:
On a fresh Fedora installation selinux seems to prohibit docker from accessing mounted files from disk.

I've added a one-line command to README as a quick solution.

There is a way to disable selinux context directly from docker-compose files but I don't know if it's not gonna impact windows users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,9 +140,8 @@ When using `selinux` (e.g. on Fedora) you may encounter an error that a mounted 
 Example error:
 
 ```
-$ docker-compoe up
+$ docker-compose up
 Starting site_site_1 ... done
-Attaching to site_site_1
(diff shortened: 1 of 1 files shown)


MiyamotoAkira:
Oh, that was the issue I had. Only appeared when using incremental.

README.md
```diff
@@ -133,6 +133,35 @@ docker system prune -a
...
 After this the next docker operation will download the correct image.
 
+## Problem with files permissions in docker (linux)
+
+When using `selinux` (e.g. on Fedora) you may encounter an error that a mounted file cannot be accessed from a docker container.
+
+Example error:
+
+```
+$ docker-compoe up
```
MiyamotoAkira: s missing

PR #1387, opened 2019-07-15
Adding moonpig case study

MiyamotoAkira:


alfredodev:
Maybe we can have two separate projects for this? the refactoring/delivery and the coaching?
What do you think?

LL782:
Over six months old. I think we should close the PR.

PR #1413, opened 2019-08-01
added details for gregory to team page

codurance-greg-cann:
Modified the team.yml to include a new entry for gregory cann and included the necessary asset. 

Verified locally using local docker build.


codurance-greg-cann:
Closing as I now have access to the codurance org, a newer PR will be opened with proper checks.

PR #1416, opened 2019-08-01
Added enric bio and photo

enricjorda:


keith-smale (changes requested):
As mentioned in Slack, suggested tweaks to the bio to adjust how it reads.
Also will need to resolve the new conflicts from merges of other branches.

enricjorda:
Corrections on the bio text added.

PR #1427, opened 2019-08-20
Seven testing sins - blog post

upsd:
I am looking for some early feedback on this.

Please ignore that the metadata on the post etc are incorrect.

It would be great to understand how easy it is to follow the post.


keith-smale:
I could follow your explanations of each sins quite well. 

Had you thought about putting numbers in front of each or your testing sins? Where the article title is that there are seven thought it would be nice to show each with it's number, just a thought.

For me the summary seems quite short so not sure if there is more you can add to provide some overview of the methods you mentioned in the article to refactor tests.
Or remove the summary title altogether and just add some spacing between the final sin and your final line about refactoring and the quote.

upsd:
> I could follow your explanations of each sins quite well.
> 
> Had you thought about putting numbers in front of each or your testing sins? Where the article title is that there are seven thought it would be nice to show each with it's number, just a thought.
> 
> For me the summary seems quite short so not sure if there is more you can add to provide some overview of the methods you mentioned in the article to refactor tests.
> Or remove the summary title altogether and just add some spacing between the final sin and your final line about refactoring and the quote.

I agree with the suggestion about removing the `summary` title.

PR #1445, opened 2019-09-18
Add ade profile

midmandle:
Added a profile with picture for Ade. Can someone confirm Ade's customer visible role for me? Is it still 'Software Craftsperson' or do we have an equivalent for QA roles?


Mattsi-Jansky (changes requested):
The aspect ratio of the photo is wrong. It needs to be the same as all of the other photos otherwise it messes up the layout. The picture demonstrates the issue. Other than that it looks good
image

src/_data/team.yml
```diff
@@ -870,4 +870,15 @@
...
-  office: manchester
\ No newline at end of file
+  office: manchester
+
+- id: ade_adegboye
+  name: Ade Adegboye
+  image: /assets/custom/img/team/ade_adegboye.jpg
+  role: Software Craftsperson
+  fullDescription:
+    <p>Ade is someone who like to experiment with things. He likes to try and do things in different ways if the normal approach does not seem to work out so well.</p>
```
keith-smale: I think this should 'likes' rather than 'like'.

keith-smale:
@steve-codurance are you able to answer Alasdair's comment re the name of Ade's role for the team page?

PR #1455, opened 2019-10-07
Removes SC London 2019 banner as well as JS only used by banner from page

jordanrobinson:
Note I've left the files in place, as I'm assuming we'll want to reuse this banner ad capability, if not for SC London 2020 then for something else.


keith-smale:
Looks good to me.

PR #1470, opened 2019-10-11
Elastic stack introduction blog post

ghost:


src/_posts/2019-10-11-elastic-stack-introduction.md
```diff
@@ -0,0 +1,915 @@
...
+image:
+ src: /assets/custom/img/blog/2019-10-11-elastic-stack-introduction/monitor.jpg
+tags:
+- monitoring
+- logging
+---
+
+# Introduction
+
+But when the software that you are running is critical for your company you can't have logs just for analysing things that happened some time ago, having the client tells you that your application is broken while you didn't even know what was happening is a real problem, and one of the ways to deal with that is using monitoring and logging.
```
pierre-jean: Why does this sentence start with `But` ? It seems to follow a previous statement but is the first sentence of the post.

src/_posts/2019-10-11-elastic-stack-introduction.md
```diff
@@ -0,0 +1,915 @@
...
+tags:
+- monitoring
+- logging
+---
+
+# Introduction
+
+But when the software that you are running is critical for your company you can't have logs just for analysing things that happened some time ago, having the client tells you that your application is broken while you didn't even know what was happening is a real problem, and one of the ways to deal with that is using monitoring and logging. 
+
+Most applications will have logging, even if is to a text file that later someone will use tools like `grep` or `tail` to see what's going on, but that's a very passive way to see the logs, you have to take action, and probably you will only look at the logs when there's something wrong. What you need is to have your logs telling you what is happening, and today we are going to build the first step for that.
```
pierre-jean: > Most applications will have logging, even if is to a text file that later someone will use tools like `grep` or `tail` to see what's going on, but that's a very passive way to see the logs, you have to take action, and probably you will only look at the logs when there's something wrong

Maybe could be rephrase:

> Most applications will have logging, even if it is just a text file that will require the use of tools like `grep` or `tail` to see what's going on. But that's a very passive way to see the logs, as you have to take action, and probably you will only look at the logs when there's something wrong

pierre-jean (commented):
Just had a look at the introduction, will review rest later ;)

keith-smale:
The main image you are using for the post is a portrait image, which doesn't work when shown on the landing page of the site (see screenshot). 
I think you could probably crop the image to make it a landscape image instead and it would probably then work ok.

Screenshot 2019-10-15 at 22 15 47


Also if you look at the snippet of text shown for you post in the screenshot, because you have an 'Introduction' heading this is shown in that snippet which doesn't make it read very well from looking at it on the landing page.
I would suggest removing that Introduction heading completely.

I've started reading the post but i'll carry on tomorrow

PR #1474, opened 2019-10-11
New clients

anguspaterson:
Add new clients


keith-smale:
Looks good, although I think we should have some copy about how we have helped each client, like we have for the other clients before merging.

keith-smale:
Is there a reason why you've included the video hero banner commit from your other PR's in this branch? 

Personally, I would have expected to just see the changes to the clients section as per the PR and branch descriptions rather than changes to other sections from other PR's.
So the commits, branch and PR should relate to the change you are making.

keith-smale:
We also have some clients listed on the 'Services' section. Should we adjust that section too?

anguspaterson:
@keith-smale @davidgrahamhall - This branch got messy and appreciate there are some changes. Will amend and setup a new branch for these changes.

PR #1475, opened 2019-10-11
ADD video hero banner on homepage.

anguspaterson:
- Add additional media query to handle extra large screens
- Create fallback pure CSS carousel for mobile devices
- Not tested on ie11, please test if possible


keith-smale:
Really nice. Like it.

anguspaterson:
Thanks! Glad to hear.

LL782 (approved):
Looks awesome in Safari at all breakpoints (haven't tested in other browsers)

There's an unmentioned change that stops the blog posts appearing in the `Activity panel` but as discussed offline this is intentional.

Let's get this in!

PR #1477, opened 2019-10-11
Podcast image update

anguspaterson:
Update thumbnail with Codurance branded image


MiyamotoAkira:
Can't say I like that image for the podcast. Is pretty much boring. The prairie dog has more personality. Could be improved? Yeah, maybe a variation based on it can be done, rather than just replace with a microphone.

keith-smale:
@MiyamotoAkira what about 😄 
Screenshot 2019-10-12 at 08 23 43

keith-smale:
This PR looks to include all the changes for the other open PR's that you submitted. 

I think if you were to make changes relating to one context in one branch then it would be easier to follow. For example, this PR is for updating the podcast image in each podcast but the actual adding of the image file was also covered in a previous PR. 
So if those two commits were together in this branch I could see those two changes together.

anguspaterson:
Agreed. Apologies for the confusion. Coming from a design background GIT workflows are relatively new to me so I'm learning as I go. All feedback appreciated and looking forward to honing my skills.

anguspaterson:
Thanks for your comments @MiyamotoAkira and I appreciate where you are coming from. This initial tweak is a temporary measure to get something in place that aligns well with the other image tweaks we did to the latest video and Kata. Moving forward I'd like to propose we take some time to discuss how best to categorise podcasts and then apply a vibrant look that ensures no loss of personality.

anguspaterson:
@MiyamotoAkira @keith-smale Let's park this for now and get together to come up with something everyone is happy with.

PR #1479, opened 2019-10-14
ADD new event to showcase Kata on homepage within highlights section

LL782:
ADD new event to showcase Kata on homepage within highlights section


anguspaterson (approved):
Thanks for cherry-picking this commit - looks good to me

PR #1480, opened 2019-10-14
Remove google map and contact details

anguspaterson:
Remove google map and contact details from 'Our Story' page. This is due to this content now being available on the contact page.


LL782 (approved):
Looks like it does what it says on the description. Not QA'd it myself

PR #1481, opened 2019-10-14
Update thumbnail with a new image of Sandro

anguspaterson:


LL782 (commented):
For me the new image isn't as good a representation of the post as the current one

LL782 (approved):
Much prefer this one! 

Is the PNG used anywhere else? If not, let's delete it

PR #1490, opened 2019-10-18
Add new clients

anguspaterson:
This commit adds new clients to the site. I appreciate that that in some cases the description does not include a reference to Codurances involvement in the project. This is an initial deploy to ensure the site is up to date with our roster of clients and descriptions to be iterated on in due course.


anguspaterson:
Is this ok to merge?

keith-smale (changes requested):
The cxloyalty logo image doesn't seem to have been added so this will need to be added.

keith-smale:
Still not really keen that there isn't anything in the captions about what Codurance is doing for these clients. If we add this as it is how long will it be until we have some text providing some details as to what we are doing for these new clients? 
@davidgrahamhall @stevelydford do you have some details as to how we are helping these clients that we can add?

Also are there any new BCN clients that we should be adding too?

keith-smale:
Also not something done by this branch, but if you compare the logos on the landing page with the logos on the client page, the ones on the landing page look a little squished.
Not sure if it's an easy fix.

anguspaterson:
Not sure why image the cxLoyalty image did not upload but done now. I will take a look at the logo squish and add the fix on this branch

keith-smale:
I spoke with David today and he came up with some additional details about what Codurance is doing for our new clients which can be added to complete this branch.

anguspaterson:
I've made these changes on the branch and pushed. Not sure if I need to do anything else?

keith-smale:
I can't see the additional captions for each client have been added. David came up with these did he send them to you?

anguspaterson:
These changes should be here now.

keith-smale (approved):

PR #1498, opened 2019-10-22
Align brand colour

anguspaterson:
Changes to add existing classes to markup and introduced new overrides to ensure the correct brand orange hex (EF772) is being used throughout the site. In some instances, a variant off-brand orange was being referenced from CSS set in the unify-2.4 theme.


keith-smale:
@anguspaterson you seem to have included your other commits relating to adding the new client logos along with the colour alignment change.

Will need to pick the last commit into a new branch so it can be merged on it's own.

anguspaterson:
I will close this request and set up a new one. Still getting to grips with this

PR #1504, opened 2019-10-24
Colour Alignment

anguspaterson:
Review and correct all instances of colour inconsistency across the site using overrides

### DIFF
--- a/src/_includes/publications.html
+++ b/src/_includes/publications.html
@@ -43,10 +43,10 @@
                         {% include post.html post=post %}
                     {% elsif forloop.index0 == 2 %} 
                         {% include post.html post=post %}
-                        <div class="masonry-grid-item col-sm-6 col-lg-4 g-mb-30 Interested in joining us">
+                        <div class="masonry-grid-item col-sm-6 col-lg-4 g-mb-30">
                             <!-- Static Quote -->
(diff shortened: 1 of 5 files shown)


src/_includes/publications.html
```diff
@@ -33,20 +33,19 @@
...
+                                <h3 class="g-color-white-opacity-0_8 g-font-size-13 text-uppercase">Try Katalyst</h3>                
                                 <a class="u-link-v2" href="http://katalyst.codurance.com"></a>
                             </article>
                             <!-- End Static Quote -->
                         </div>
                         {% include post.html post=post %}
                     {% elsif forloop.index0 == 2 %} 
                         {% include post.html post=post %}
-                        <div class="masonry-grid-item col-sm-6 col-lg-4 g-mb-30" >
+                        <div class="masonry-grid-item col-sm-6 col-lg-4 g-mb-30 Interested in joining us">
```
ghost: Not sure that "Interested in joining us" is what we want here 🤔

anguspaterson: Very true! This was a mistake. It's been updated on this branch. Good spot!

keith-smale: Did you push the change? I can't see it's been updated.

PR #1508, opened 2019-10-25
Add youtube to social icons

### LINKED ISSUE #1514: Add youtube link to the footer of the webpages.
As described in the title.

The link is this: https://www.youtube.com/user/codurance

keith-smale:
Added YouTube icon to the social icons in the footer.
Also added target="_blank" to all the social icons so when people click a new tab is opened rather than re-directing away from the Codurance site.

The icon looks a bit rubbish given the size of the icons and what we have to work with from the icon set in the unify templates.
Maybe making all the icons a little bigger may help with the look.

### DIFF
--- a/src/_includes/footer.html
+++ b/src/_includes/footer.html
@@ -154,22 +154,22 @@
             </a>
           </li>
           <li class="list-inline-item g-mx-10" data-toggle="tooltip" data-placement="top" title="Facebook">
-            <a href="https://www.facebook.com/codurance/" class="g-color-white-opacity-0_5 g-color-white--hover">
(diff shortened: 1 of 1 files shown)


anguspaterson:
Agreed that this looks a bit small I am looking at how best to resize without breaking the layout on smaller screen sizes. Good work on the _target="blank" much better.

tonycodurance:
> Agreed that this looks a bit small I am looking at how best to resize without breaking the layout on smaller screen sizes. Good work on the _target="blank" much better.

👋 hey! Any update on this one? Are we happy to deploy?

keith-smale:
No longer needed as already done.

PR #1511, opened 2019-10-28
Add Eloy to Team

eloyacostapadron:


src/_data/team.yml
```diff
@@ -1000,3 +1000,15 @@
...
   office: london
+
+- id: eloy_acosta
+  name: Omar Eloy Acosta Padron
+  image: /assets/custom/img/team/eloy_acosta.jpg
+  role: Software Craftsman
+  fullDescription:
+    <p>Eloy is constantly improving personal and professional skills for providing quality software solutions and systems.</p>
+    <p>His first contact with computers was in his childhood with the Epson QX 10.</p>
+    <p>Since obtained a degree of Licentiate in Computer Engineering by the ULPGC, he has been involved in managing and developing projects for several companies using a wide variety of technologies and practices.</p>
```
keith-smale: This part doesn't read quite right. It should be 'Since **he** obtained a degree' or Since **obtaining** a degree'.

keith-smale (changes requested):
Thank you for adding yourself. You didn't need any help!!!!

Just one small change then it will be all good. 
There is another change pending from someone else to the team.yml, so there may be some conflicts once you make your change, so if there are conflicts please resolve those too.

Thanks

src/_data/team.yml
eloyacostapadron: Thanks Keith for looking at this.
I have update to correct the text.

keith-smale (approved):

keith-smale:
The changes are all good. Just need you to resolve the conflicts and it's good to go.

PR #1512, opened 2019-10-28
Align logo colours and fix broken link on training page

anguspaterson:
Small fixes to logo and correction of broken link


anguspaterson:
The logo has been updated to ensure the red yellow and orange are correct. We have decided to retain the text as black as opposed to teal so as to retain consistency with the rest of text in the header which is a shade of the black.

PR #1515, opened 2019-10-29
Add myself to the team page

ghost:


keith-smale:
@smeny-codurance Looks as though you have forked off the Codurance repo rather than cloning the repo and making your changes there.

The CI doesn't run when changes from a forked repo are made.

Can you clone the site repo and create a branch to add yourself and we can get you added that way.

Thanks

PR #1527, opened 2019-11-17
Changed how the front page content is calculated

MiyamotoAkira:
This way we go back to showing blogs, and we can make videos/podcasts
to appear separately


anguspaterson:
Looks good to me. I am going to implement a fix on another request to ensure that the imagery on single blog post headers have a darker overlay as I've noticed a legibility issue on the latest post.

PR #1535, opened 2019-11-28
Replace Socrates with SC London on homepage

anguspaterson:
Switch out Socrates Teaser to SC London. New Image from SC 2019, text taken from https://sc-london.com/

Target set to _blank as it links to https://sc-london.com/


keith-smale (commented):
Is Socrates being removed completely? If so we should probably remove the english translation `index-software_craftsmanship-socrates` too.

LL782:
@keith-smale @anguspaterson is this pull request still wanted? Can we close it or make it mergeable?

PR #1536, opened 2019-11-28
Add 'See More' CTA to client logos block

anguspaterson:
Add see more CTA to client logos block on the home page. This links through to the clients page.


keith-smale:
Looks good.

keith-smale (approved):

PR #1543, opened 2019-12-16
Podcast 20 entry

MiyamotoAkira:


keith-smale:
@MiyamotoAkira the podbean player is smaller than the previous players on other podcasts. I prefer the smaller one. 
Seems there is a lot of empty space under the player which looks to be some styling connected to the players iframe getting some fixed height.
Not sure if that can be adjusted.

PR #1548, opened 2019-12-30
Good Bye Folks👋

archel:


chriseyre2000:
Will need to manually remove the image from the S3 bucket when this is merged.

PR #1575, opened 2020-01-20
adjust youtube iframe width to be responsive

keith-smale:
Updated iframe width on test driving CDN post so it fits the screen on mobile and smaller devices.


mattgraygithub (commented):
Looks good to me

PR #1591, opened 2020-02-03
modify the position of the apprentices to Software Craftsman

josewenzel:


ghost:
LGTM but why not using the gender neutral titles instead?
(After all this is how it appears on our work contract)

PR #1595, opened 2020-02-05
Redirect /fizzbuzz to Katalyst

jaramir:


Mattsi-Jansky:
In preview it only works if you visit `index.htm`, visiting `http://codurance-site-pr.s3-website-eu-west-1.amazonaws.com/site-redirect-fizzbuzz/en/fizzbuzz` doesn't work. Presumably this is only because the S3 bucket is configured differently, and it'll work in production right?

jaramir:
> it'll work in production right?

🤞

Mattsi-Jansky (approved):
Let's roll the 🎲

jaramir:
> Let's roll the 🎲

nope, the /pubications redirect works on the preview so /fizzbuzz does not work for some reason

Mattsi-Jansky:
Ah, good spot.

jaramir:
wrong file extension 😄

Mattsi-Jansky (approved):
Good stuff, works now 👍

PR #1619, opened 2020-02-19
add posting guide

josewenzel:


sherlockq:
Very well written! What about move it to the project root so people don't have to know where to find the README first.

keith-smale:
There is a section in the ReadMe called 'Flight Rules'. Is that part redundant with your Posting Guide? 
We probably should try and merge the Flight Rules section with your Posting Guide section.

keith-smale:
Thanks for making the update. Looks all good to me.

PR #1620, opened 2020-02-19
Sort author's publications descendingly

### LINKED ISSUE #1577: Blog Post ordering on Authors Page
When you navigate to the blog posts or articles for a specific Author the posts are shown oldest to newest. 

To me it would make more sense to show the newest posts first.

**Info from Mattsi** 
The fix would have to go in author_generator.rb:47, around here: self.data['publications'] = site.collections.values.flatten.select do |post| post.data["author"] == author end
Just needs to be piped into some sorting function.

sherlockq:


spec/_plugins/alias_generator_spec.rb
```diff
@@ -28,7 +28,35 @@
       })
     
     author_index = Jekyll::AuthorIndex.new(site, "base", "author_dir", "Author") 
-    expect(author_index.data).to eq({ 'author' => "Author", 'description' => 'author: Author', "publications" => [video, post], 'title' => "Author"})
+    
+    expect(author_index.data['author']).to eq("Author")
+    expect(author_index.data['description']).to eq("author: Author")
+    expect(author_index.data['publications']).to include(video, post)
+    expect(author_index.data['publications']).not_to include(newsletter)
+    expect(author_index.data['title']).to eq("Author")
```
Mattsi-Jansky, line 36: I think the previous format of assertion (ie, comparing the entire object in one go rather than properties of the object in multiple lines) was better as it expresses the same amount and more in one line.

spec/_plugins/alias_generator_spec.rb
```diff
@@ -28,7 +28,35 @@
...
+
+    allow(Jekyll::Utils).to receive(:merged_file_read_opts)
+    allow(File).to receive(:read).and_return(file)
+
+    allow(site).to receive(:in_source_dir)
+    allow(site).to receive(:file_read_opts)
+    allow(site).to receive(:config).and_return({})
+    
+    allow(early_post).to receive(:data).and_return({ 'author' => "Author", 'title' => "", 'date' => Date.new(2000,1,1) })
+    allow(late_post).to receive(:data).and_return({ 'author' => "Author", 'title' => "", 'date' => Date.new(2001,1,1) })
```
Mattsi-Jansky: The "early" and "late" naming confused me a bit; "older" and "newer" may be better. But, that could just be me.

src/_plugins/author_generator.rb
```diff
@@ -44,8 +44,11 @@ def initialize(site, base, author_dir, author)
       # Set the meta-description for this page.
       meta_description_prefix  = site.config['author_meta_description_prefix'] || 'author: '
       self.data['description'] = "#{meta_description_prefix}#{author}"
-      self.data['publications'] = site.collections.values.flatten.select do |post| post.data["author"] == author end
-      
+      # Posts should be sorted by date descendingly
+      publications = site.collections.values.flatten.select do |post| post.data["author"] == author end
+      publications = publications.sort_by{ |post| post.data["date"] }.reverse
+      self.data['publications'] = publications
```
Mattsi-Jansky: It's probably possible to do this without multiple assignments, something like;

```
self.data['publications'] = site.collections.values.flatten.select 
  do |post| post.data["author"] == author end 
  .sort_by{ |post| post.data["date"] }.reverse
```

It all depends on how Ruby handles newlines, not sure how it does that.

Mattsi-Jansky (approved):
Looks great! Thanks for doing this man. Left a few comments, but nothing significant and nothing worth preventing merge.

keith-smale:
I agree with Mattsi on the early, late naming. New, Old1 is better and makes more sense.

Looks like it work great. Good stuff.

spec/_plugins/alias_generator_spec.rb
sherlockq: Nice suggestion!

src/_plugins/author_generator.rb
sherlockq: Worth a try, I'll try write it as a flow way.

spec/_plugins/alias_generator_spec.rb
sherlockq: This is debatable. For several reasons I prefer the modified version
1. Avoid the concern of sorting in this test case
2. Easy to read, especially the include and not include.
3. No implicit assumption of the eq operation's implementation. May not be a thing for Ruby, but quite a concern for Java.

Mattsi-Jansky: That's fair, I think it comes down to personal taste.

sherlockq:
Yeah mostly personal taste. Also it depends what are intended to be tested
in one test case.
The old one will break if a new field is introduced, which seems brittle to
me. So after some troubles had before I just switched to the current way.

Zhiqiang (Sherlock) Qiao
Software Craftsman @ Codurance

keith-smale:
@sherlockq Can you update the early/late part to older/newer then we can get this merged.


Thanks

sherlockq:
Since it's a trival change, I'll fix it and just merge it.

PR #1628, opened 2020-02-20
Change ogp image

### LINKED ISSUE #1038: OGP embed image is too big
When we embed the OGP image (in this case using [Facebook's OGP Debugger](https://developers.facebook.com/tools/debug/sharing/?q=https%3A%2F%2Fcodurance.com%2F)) it comes up as such: 

image

See: https://opengraphcheck.com/result.php?url=https%3A%2F%2Fcodurance.com%2F

As you can see, it's too big and only a small part of cut-off text is appearing. I recommend changing ti . to `https://codurance.com/assets/custom/img/codurance-logo-artwork.png`, which is the smaller logo-only image we use for our Twitter-specific embed image. That comes out like so: 

image

mattgraygithub:
Resolves [issues/1038](https://github.com/codurance/site/issues/1038). [Facebook](https://developers.facebook.com/tools/debug/?q=codurance.com) seem to have changed their ogp settings since the issue was raised, but the image was still the wrong size. So I've added a new image of the correct size.


mattgraygithub:
Merged it myself as it's only a very small change and Sherlock checked it out and gave a second opinion

PR #1633, opened 2020-02-21
remove trainings that we no longer do

josewenzel:


src/_data/trainings_en.yml
```diff
@@ -33,22 +33,3 @@
...
-  page: /services/training/service-fabric-training
-
-- id: docker
-  name: Zero to Continous Delivery with Docker
-  author: Robert Firek
-  description: Learn to use Docker tools to enable continuous deployment, horizontal scalability & zero downtime.
-  page: /services/training/devops-training
-
-
-- id: crafting_webapps
```
threenary, line 50: I won't remove this training, we are actually re-building it for a Barcelona client. Please correct me if I am wrong @shadowman 
I imagine we should update the author @JordiCreus

upsd:
Looks legit 👍 

Thanks for doing this.

upsd (approved):
LGTM

threenary:
Why you merged this ignoring my above advice regarding the **Crafting Web Apps training**?
I would have expected some feedback from @shadowman or @JordiCreus before merging.

Unless I missed some urgent requirement to remove specifically this training I was not aware of.

src/_data/trainings_en.yml
ghost: Once I've done the training we can add this again.

PR #1631, opened 2020-02-21
Testing for S3 file removal

mattgraygithub:

PR #1637, opened 2020-02-25
Duplicate the codurance logo for mail signature

sherlockq:
It's better to have a dedicated image for signature rather than reuse an existing one in case any change in the future.

After this one is deployed every Codurancer should be notified to change the mail siganature image.

The story:
Most people just copied the mail signature from others and it turns out they are all referencing a never version controlled image (http://codurance.com/assets/img/codurance.png). After #1631 now the S3 bucket is fully synchronized with repo and that file is removed.

For now I manually uploaded the image to its old place. That file will be removed automatically at next deployment. So I'll upload that again after this one is merged.


sherlockq:
Due to the creation of #1641, this PR is closed.

PR #1642, opened 2020-02-26
added info about adding links to posts in readme

keith-smale:
Small addition to the readme to use HTML for links in blog posts so that we can take advantage of target="_blank".


ghost (approved):
Nice!

PR #1660, opened 2020-03-04
Get back mistakenly removed client projects

sherlockq:
Follow the case study links at https://codurance.com/clients/ to test it.


mattgraygithub:
Fixed

PR #1670, opened 2020-03-22
Blogpost about types part 2

ghost:


src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,524 @@
...
+image:
+    src: /assets/custom/img/blog/jigsaw-types.jpg
+    attribution:
+        text: Photo by Gabriel Crismariu on Unsplash
+        href: https://unsplash.com/photos/sOK9NjLArCw
+tags:
+    - types
+    - java
+    - kotlin
+abstract: Let's continue our conversation about types and see how to apply then on the day to day
```
nicholemellekas, line 17: add description

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,524 @@
...
+---
+
+We already spoke about the different type systems and how they work, now it's time to write some code and see
```
nicholemellekas: link to previous blog

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,524 @@
...
+{
+    "startDate": "10/11/2019",
+    "endDate": "15/11/2019",
+    "origin": "LHR",
+    "dest": "DUB"
+}
+```
+
+Now that we receive that request, we have to understand what composes a search:
+
```
nicholemellekas, line 38: This whole paragraph needs to be reformatted. Please make the terms lower case. 

- The start and end dates
- The start date has to be earlier than the end date.
- An origin and destination
- The origin and the destination must be different.
- The origin and the destination must be valid IATA

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,524 @@
...
+    - kotlin
+abstract: Let's continue our conversation about types and see how to apply then on the day to day
+---
+
+We already spoke about the different type systems and how they work, now it's time to write some code and see
+how type can help us.
+
+## The Flight Search Example
+
+Imagine that we are building a company that searches for flights on multiple websites, we are exposing an endpoint that accepts JSON. Right now we are only dealing with simple searches where all flights will have a return and the accepted JSON is:
```
nicholemellekas: This should be two sentences. 

Imagine that we are building a company that searches for flights on multiple websites. We are exposing an endpoint that accepts JSON.

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,524 @@
...
+
+Now that we receive that request, we have to understand what composes a search:
+
+- Start and End dates
+- The Start date has to be earlier than the End date.
+- An Origin and Destination
+- The Origin and the destination must be different.
+- The Origin and Destination must be valid IATA
+
+We can have all those validations without creating a class, imagine that we have a controller that will receive that, parse the JSON and send to a service.
```
nicholemellekas: Reformat:

We can have all those validations without creating a class. Imagine that we have a controller that will receive that, parse the JSON, and send to a service.

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,524 @@
...
+    		return searchService.findFlights(
+    			searchObject.get("startDate").asText(),
+    			searchObject.get("endDate").asText(),
+    			searchObject.get("origin").asText(),
+    			searchObject.get("dest").asText());
+    	}
+    } 
+```
+
+Now all the validations that we have to do must live inside the `SearchSevice`, not only the business validations, we must do validations that are related to application code like date formatting, but not only that, in this way inside our class we will have to deal with multiple parameters that we will have to pass them together every time we need a `SearchRequest`.
```
nicholemellekas: Rework, this sentence isn't clear. It should be broken into multiple sentences. 

Now all the validations that we have to do must live inside the `SearchSevice`, not only the business validations, we must do validations that are related to application code like date formatting, but not only that, in this way inside our class we will have to deal with multiple parameters that we will have to pass them together every time we need a `SearchRequest`.

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,524 @@
...
+```java 
+class SearchRepository {
+
+    public List<Flight> findFlights(String startDate, String endDate, String origin, String dest) {
+        // implementation
+    }    
+}
+```
+
+There are so many smells in that snippet that made me sick, jokes aside we have to see that we are moving all the validations to the edge of the application, this will only blow up when we make a database call. This might be a problem for error handling because we want to tell the person that called the API which kind of error is, a database problem would usually be a `5XX`, but in reality, could be a `4XX` since the problem is in the payload that was sent, not in the database.
```
nicholemellekas: Reformat: 

There are so many smells in that snippet that made me sick. Jokes aside, we have to see that we are moving all the validations to the edge of the application. This will only blow up when we make a database call. This might be a problem for error handling because we want to tell the person that called the API which kind of error is, a database problem would usually be a `5XX`, but in reality, could be a `4XX` since the problem is in the payload that was sent, not in the database.

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,525 @@
...
+        // implementation
+    }    
+}
+```
+
+There are so many smells in that snippet that made me sick, jokes aside we have to see that we are moving all the validations to the edge of the application, this will only blow up when we make a database call. This might be a problem for error handling because we want to tell the person that called the API which kind of error is, a database problem would usually be a `5XX`, but in reality, could be a `4XX` since the problem is in the payload that was sent, not in the database. 
+
+### Fastening the type seatbelt
+
+Let's start with the dates, you know there are not multiple date format and isn't a problem that we face frequently. We have to date fields that are going to throw an exception just when we execute the query, the problem is that is already too late, we don't want them to reach the database.
```
nicholemellekas: Let's start with the dates, there are not multiple date formats and this isn't a problem that we face frequently.

nicholemellekas: Unclear: 

We have to date fields that are going to throw an exception just when we execute the query, the problem is that is already too late, we don't want them to reach the database.

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,525 @@
...
+ 
+    private static LocalDate parseDate(String date) {
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("dd/MM/yyyy");
+        return LocalDate.parse(date, formatter);
+    }
+
+}
+```
+
+Now with the parser being done in the service we replace exceptions related to our database for `DateTimeParseException`, this makes way easier to capture the right exception instead of trying to figure it out what was happening. What we have now is better than the previous code using strings all around but we can and must do better. The `SearchService` is throwing `DateTimeParseException` and we can handle that case in the controller and return something like `400 - Bad Request`.
```
nicholemellekas: Now with the parser being done in the service we replace exceptions related to our database for `DateTimeParseException`, this makes it way easier to capture the right exception instead of trying to figure it out what was happening.

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,525 @@
...
+```java
+class SearchRepository {
+
+    public List<Flight> findFlights(LocalDate startDate, LocalDate endDate, IATA origin, IATA dest) {
+        // implementation
+    }
+}
+```    
+
+With those changes, we can at least guarantee that the dates passed to the database are valid and the formatting don't matter much at this point now because it's encapsulated inside of a class. We can't confirm that the date exists in the database but there won't be any exceptions thrown when the query is executed.
```
nicholemellekas: With those changes, we can at least guarantee that the dates passed to the database are valid and the formatting doesn't matter much at this point now because it's encapsulated inside of a class.

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,525 @@
...
+
+    public List<Flight> findFlights(LocalDate startDate, LocalDate endDate, IATA origin, IATA dest) {
+        // implementation
+    }
+}
+```    
+
+With those changes, we can at least guarantee that the dates passed to the database are valid and the formatting don't matter much at this point now because it's encapsulated inside of a class. We can't confirm that the date exists in the database but there won't be any exceptions thrown when the query is executed. 
+
+The state of the services and repositories are improving, said that there's still quite some work that can be done. We are doing all the validations inside the SearchService.
```
nicholemellekas: Unclear

The state of the services and repositories are improving, said that there's still quite some work that can be done

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,525 @@
...
+
+The state of the services and repositories are improving, said that there's still quite some work that can be done. We are doing all the validations inside the SearchService. 
+
+Just a clarification, The Application Layer is the part that handles the communication, in this case, it would be the controller. The controller isn't related to the business but just a way of input/output for our domain. This diagram shows the boundaries between them:
+
+![Diagram of the application with the controller, flight service and flight repository]({{ '/assets/custom/img/blog/2020-03-22-types-and-accepting-the-facts/application-diagram.jpg' | prepend: site.baseurl }})
+
+We use types, dependency injection and interfaces to abstract what the application is doing. The `FlightSearchService` don't care if the data is coming through HTTP, RPC or even a CLI. The same goes for the `FlightSearchRepository`, it just cares that you can store and retrieve the data later, the how doesn't matter for the business, that's an application responsibility.
```
nicholemellekas: The `FlightSearchService` doesn't care if the data is coming through HTTP, RPC, or even a CLI.

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,525 @@
...
+
+Just a clarification, The Application Layer is the part that handles the communication, in this case, it would be the controller. The controller isn't related to the business but just a way of input/output for our domain. This diagram shows the boundaries between them:
+
+![Diagram of the application with the controller, flight service and flight repository]({{ '/assets/custom/img/blog/2020-03-22-types-and-accepting-the-facts/application-diagram.jpg' | prepend: site.baseurl }})
+
+We use types, dependency injection and interfaces to abstract what the application is doing. The `FlightSearchService` don't care if the data is coming through HTTP, RPC or even a CLI. The same goes for the `FlightSearchRepository`, it just cares that you can store and retrieve the data later, the how doesn't matter for the business, that's an application responsibility. 
+
+Continuing with the changes. Now we can see that the problem with the codebase is that is the service which is part of our business domain and we are filling with application code that has nothing to do with it. The solution for that is moving that code up the application layer.
```
nicholemellekas: Change this sentence: Now we can see that the problem with the codebase is that is the service which is part of our business domain and we are filling with application code that has nothing to do with it.

Suggested sentence: The problem with the codebase is that it's being filled with application code that isn't relevant.

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,525 @@
...
+
+All those things that I've said here aren't made up shit that I'm coming to try to look smart, they are code smells that many other people have written about before. The name of the code smell that we just changed is Primitive Obsession, it's a code smell were you use primitive types to deal with things that should be abstracted as an object. 
+
+### Throwing Exception where, when, how?
+
+We already mentioned about the Domain and the Application layer, where should we be adding the validation for the values that we have. For things like parsing dates or a JSON which is explicit out of the business domain, it's better to make them live inside the Application Layer so we can test the Business Layer without having to worry about that, also the way we drive the application might be different depending on what you want. That isn't something that our Business should be worried about.
```
nicholemellekas: We already mentioned the Domain and the Application layer,...

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,525 @@
...
+    public static void main(String[] args) {
+        String home = System.getenv("HOME");
+        System.out.println(home.length());
+    }
+}
+```
+
+This snippet will throw a NullPointerException but at any moment we were warned that the method would return null, we might know for reading the documentation or the code. We can add a null check before calling `home.length()`, that solves the problem of the exception that we are getting, we still have the problem that we have to do that after every call of the method, and human beings are unreliable to do repetitive tasks like that. Do you know who is good checking that kind of stuff? The compiler of course. 
+
+With all the advance of the modern society and Java, there's a quite decent way of dealing with this problem. Java provides us with the `Optional<>` that can wrap null values for us. The main advantage of using an Optional is that we can't use the value straight away (please, don't call `.get()` straight away), we explicitly have to deal with the possibility of a null value. This is way better than returning null or just throwing an exception. The final result would be:
```
nicholemellekas: With all the advances in modern society and Java...

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,525 @@
...
+
+// Good
+items.add(item)
+```
+
+You should also search for the methods of your language that are immutable, like in Java the Instant method is immutable but LocalDate doesn't. 
+
+During the examples, there were also many constructors with validations and more code than the usual, if you are doing that a lot you should totally learn about Static Factories that are mentioned in Effective Java, it will teach you how to write more idiomatic constructors for your classes.
+
+There's also Inline classes, that's something that is coming to the next version of Kotlin and to some future version in Java. When you need to wrap a single value like an Id. I will not give any examples but you can check those two sources for [Kotlin](https://kotlinlang.org/docs/reference/inline-classes.html) and [Java](https://www.infoq.com/articles/inline-classes-java/)
```
nicholemellekas: Add a full stop at the end of the paragraph.

src/_posts/2020-03-22-types-and-accepting-the-facts.md
```diff
@@ -0,0 +1,525 @@
...
+
+## Sources
+
+- Primitive Obsession
```
nicholemellekas: Should these be links?

PR #1678, opened 2020-04-02
Update 2019-04-16-containers-with-docker.md

ghost:
Small fix on two code snippets.

### DIFF
--- a/src/_posts/2019-04-16-containers-with-docker.md
+++ b/src/_posts/2019-04-16-containers-with-docker.md
@@ -125,7 +125,7 @@ This is probably one of the most used commands that you will encounter.
 We want to be able to delete an image, stop a container that is running, and delete a previous container. To do that we have a set of commands:
 
 - `docker ps`: Show running containers. Add the `-a` flag to show stopped containers.
-- `docker rm <contaier id or name>`: Delete a container.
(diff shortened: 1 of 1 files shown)


src/_posts/2019-04-16-containers-with-docker.md
```diff
@@ -252,7 +252,8 @@ Finally, we can run the application by passing the `CMD` to start
 
 ```Dockerfile
 FROM adoptopenjdk/openjdk11-openj9
-WORKDIR /realworldCOPY . /realworld
+WORKDIR /realworld
+COPY . /realworld
```
Mattsi-Jansky, line 256: I would do `COPY . .` so that you don't have to define the working directory in two places.

ghost: I'm just fixing the typo on the original code snippet, maybe Andre could intervene on that?

ghost: `COPY . .` will also work in this case. I don't mind having that changed.

ghost: IMHO, I think that keeping it the way you initially wrote it documents your progress as well.

PR #1689, opened 2020-04-13
Partners Skillsmatter

anguspaterson:
Added skills matter logo to the homepage partners block. 2-up side by side om mobile and 5-up on desktop


keith-smale (approved):
All good to go

PR #1690, opened 2020-04-15
Softmod styling

keith-smale:
This branch is for all changes in relation to the Software Modernisation campaign.


src/_includes/head.html
```diff
@@ -1,24 +1,43 @@
...
+    {% else %}
+        {% if page.title %}
+            {{ page.title }}
+        {% else %}
+            Software Craftsmanship and Agile Development
+        {% endif %}
+        | Codurance | Craft at Heart | London | Barcelona
+    {% endif %}
+</title>
+
```
LL782, line 15: This is starting to get messy! It's still readable just about but shouldn't this logic be in a single place wherever `page.title` is defined. 

This is not a comment to be addressed before merging (I know you guys are working super fast for some reasons.)  But outside of this PR would should talk about the website

src/_includes/head.html
```diff
@@ -1,24 +1,43 @@
...
  <!-- Meta -->
 <meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
-<meta name="description" content="Our team of dedicated software craftspeople provide consultancy, software development, and training services to clients seeking high quality development processes and software solutions.">
+<meta name="description" content="{% if page.url == '/software-modernisation/' and site.lang == 'es' %}{{ page.description-es }}{% elsif page.url == '/software-modernisation/' and site.lang != 'es' %}{{ page.description }}{% else %}Our team of dedicated software craftspeople provide consultancy, software development, and training services to clients seeking high quality development processes and software solutions.{% endif %}" />
 <meta name="author" content="Codurance Ltd">
```
LL782, line 20: Same deal here. It feels like this logic should be inside `page.description` logic rather than spread into the HTML as well.

src/_includes/head.html
```diff
@@ -1,24 +1,43 @@
...
 
+{% if page.metarobots %}
+ <meta name="robots" content="{{ page.metarobots }}" />
+{% endif %}
```
LL782, line 24: Nice to see this stuff being added 👍

src/_includes/head.html
```diff
@@ -1,24 +1,43 @@
...
+<meta name="twitter:card" content="summary_large_image" />
+<meta name="twitter:site" content="{% if site.lang == 'es' %}@codurance_es{% else %}@codurance{% endif %}"/>
+<meta name="twitter:title" content="{% if page.url == '/software-modernisation/' and site.lang == 'es' %}{{ page.title-es }}{% elsif page.url == '/software-modernisation/' and site.lang != 'es' %}{{ page.title }}{% else %}{% if page.title %}{{ page.title }}{% else %}Codurance{% endif %}{% endif %}" />
+<meta name="twitter:description" content="{% if page.url == '/software-modernisation/' and site.lang == 'es' %}{{ page.description-es }}{% elsif page.url == '/software-modernisation/' and site.lang != 'es' %}{{ page.description }}{% else %}Our team of dedicated software craftspeople provide consultancy, software development, and training services to clients seeking high quality development processes and software solutions.{% endif %}" />
 <meta name="twitter:image" content="{% if page.image.src %}{{ page.image.src | prepend: site.url }}{% else %}{{ '/assets/custom/img/codurance-logo-artwork.png' | prepend: site.url }}{% endif %}" />
```
LL782, line 41: Same deal here. It feels like `page.description` should be handling all of this logic.

src/_includes/head.html
```diff
@@ -93,6 +113,8 @@
 <link rel="stylesheet" href="{{ '/assets/custom/css/custom-navbar-or-navigation.css' | prepend: site.baseurl }}">
 <link rel="stylesheet" href="{{ '/assets/custom/css/custom-layout-colours.css' | prepend: site.baseurl }}">
 <link rel="stylesheet" href="{{ '/assets/custom/css/custom-layout-buttons.css' | prepend: site.baseurl }}">
+<link rel="stylesheet" href="{{ '/assets/custom/css/custom-softmod-feature.css' | prepend: site.baseurl }}">
+<link rel="stylesheet" href="{{ '/assets/custom/css/custom-softmod-tabs.css' | prepend: site.baseurl }}">
```
LL782, line 117: So many different CSS files. I see that this PR wasn't the start of it but we need to work on bundling these things.

src/_includes/header.html
```diff
@@ -85,3 +85,7 @@
       </div>
 </header>
 <!-- End Header -->
+
+{% if page.url == "/" or page.url == "/careers/" %}
+   {% include sm_cta_slice.html%}
```
LL782, line 90: Is `sm` short for `software-modernisation`? Will that still be obvious in the future?

src/assets/custom/css/custom-softmod-feature.scss
```diff
@@ -0,0 +1,751 @@
...
+    box-shadow: 0 1px 3px rgba(0,0,0,0.12), 0 1px 2px rgba(0,0,0,0.24);
+}
+
+/********************************
+    Helpers - Spacing
+********************************/
+
+.pad-top-none {
+    padding-top: 0!important;
+}
```
LL782, line 57: 👎 Something has gone wrong somewhere if we have to resort to this

src/assets/custom/css/custom-softmod-feature.scss
```diff
@@ -0,0 +1,751 @@
...
+    }
+
+    .g-pb-100--desk {
+        padding-bottom: 100px!important;
+    }
+
+    .g-pb-200--desk {
+        padding-bottom: 200px!important;
+    }
+
```
LL782, line 93: Really not enjoying these abstract names `g-pb-200` what?

src/assets/custom/css/custom-softmod-feature.scss
```diff
@@ -0,0 +1,751 @@
...
+/********************************
+    Helpers - Backgrounds
+********************************/
+
+.bg-slate {
+    background-color: #2B3342;
+}
+
+.bg-blue {
+    background-color:#0F4071!important;
```
LL782: I find the need to call this out as important very concerning. I'm looking forward to having more time to work together on this repo, and to understand the reasons why

src/assets/custom/css/custom-softmod-feature.scss
```diff
@@ -0,0 +1,751 @@
...
+}
+
+.bg-slate--pattern {
+    background-image: url('{{site.baseurl}}/assets/custom/img/softmod/bg-slate-pattern-2.jpg');
+    background-size: 100%;
+    background-repeat: no-repeat;
+}
+
+.bg-testimonials {
+    background:url('{{site.baseurl}}/assets/custom/img/softmod/bg-testimonials.png');
```
LL782, line 123: Not sure where this asset is. It's not live currently and I don't see it in this PR. Is this className definitely being used? `.bg-testimonials`

src/assets/custom/css/custom-softmod-feature.scss
```diff
@@ -0,0 +1,751 @@
...
+        }
+
+        .sm-cta-block__lh,
+        .sm-cta-block__rh {
+            width: 50%;
+            height: 400px;
+            float: left;
+        }
+    }
+}
```
LL782, line 752: This file is huge. One day let's bring this repo up to data so we can have smaller CSS files and bundle/include them at build time?

src/assets/custom/css/custom-softmod-feature.scss
```diff
@@ -0,0 +1,751 @@
...
+        background-size: 50% auto;
+    }
+
+    .bg-testimonials {
+        background:url('{{site.baseurl}}/assets/custom/img/softmod/bg-testimonials.jpg');
+        background-size: 110% auto;
+        background-position-y: 360px;
+        background-repeat: no-repeat;
+    }
+}
```
LL782, line 184: `background-repeat` gets repeated a lot. I also don't understand why the `background-positioning` and `background-size` has to be so different for such seemingly similar elements.

src/assets/custom/css/custom-softmod-feature.scss
```diff
@@ -0,0 +1,751 @@
...
+
+    .bg-risk-management {
+        background-image: url('{{site.baseurl}}/assets/custom/img/softmod/bg-risk-management.svg');
+        background-position: center bottom 50px;
+        background-repeat: no-repeat;
+        background-size: 50% auto;
+    }
+
+    .bg-testimonials {
+        background:url('{{site.baseurl}}/assets/custom/img/softmod/bg-testimonials.jpg');
```
LL782, line 179: Looks like the shorthand `background` is being used along side longhand attributes like `background-size` etc

src/assets/custom/css/custom-softmod-feature.scss
```diff
@@ -0,0 +1,751 @@
...
+.bg-testimonials {
+    background:url('{{site.baseurl}}/assets/custom/img/softmod/bg-testimonials.png');
+    background-size: cover;
+    background-repeat: no-repeat;
+}
+
+.bg-compass-cta {
+    background:url('{{site.baseurl}}/assets/custom/img/softmod/bg-compass-cta.jpg');
+    background-size: cover;
+    background-repeat: no-repeat;
```
LL782, line 131: Weird mix of short and longhand `background` attributes

src/assets/custom/css/custom-softmod-feature.scss
```diff
@@ -0,0 +1,751 @@
...
+********************************/
+
+.z-index-1 {
+    z-index: 1;
+    position: relative;
+}
+
+.p-relative {
+    position: relative;
+}
```
LL782, line 235: There are a lot of "helpers" that are clearly available for use across multiple components. This moves the responsibility of styling into the HTML because that's where these "helpers" are applied. 

Let's think about moving towards a methodology where styles and the markup they apply to live closely together.

src/assets/custom/css/custom-softmod-feature.scss
```diff
@@ -0,0 +1,751 @@
...
+    transition: all 0.3s ease;
+
+    &:hover {
+        background-color:rgba(251, 114, 104, 0.8)!important;
+        transition: all 0.3s ease;
+    }
+}
+
+.btn--blue {
+    background-color:rgba(15, 64, 113, 1)!important;
```
LL782: Really feels like we should go to the source of the problem rather than just adding more and more overrides to button styles.

src/assets/custom/css/custom-softmod-feature.scss
```diff
@@ -0,0 +1,751 @@
...
+}
+
+.hs-form-58727484-8758-489d-acfa-9a4956308ce6_6145720f-1db3-4a25-a9e5-3edb3baa1ecf .hs-main-font-element, .submitted-message.hs-main-font-element {
+    border: 1px solid #fb7268!important;
+    text-align: center!important;
+    width: 94%!important;
+    display: block!important;
+    padding: 10px!important;
+    margin-bottom: 20px!important;
+}
```
LL782, line 317: Unreadable names and tonnes of overrides

src/assets/custom/css/custom-softmod-feature.scss
```diff
@@ -0,0 +1,751 @@
...
+.sm-hero__grid__grid-item--intro div {
+    margin: auto;
+    color: #FFFFFF!important;
+    padding-top: 50px;
+    padding-bottom: 50px;
+}
+
+.sm-hero__grid__grid-item--intro div p {
+    color: #FFFFFF!important;
+}
```
LL782, line 420: I'm wondering why we are targeting generic elements `div` `p` etc. rather than applying class names or CSS directly to the elements 🤔

src/assets/custom/css/custom-softmod-feature.scss
```diff
@@ -0,0 +1,751 @@
...
+
+.sm-hero__grid__grid-item--intro-image {
+    display: none;
+    overflow: hidden;
```
LL782, line 424: Surely we don't need this if it is `display: none`
```suggestion
```

src/assets/custom/css/custom-softmod-feature.scss
```diff
@@ -0,0 +1,751 @@
...
+
+.sm-hero__grid__grid-item--intro-image img {
+    width: 110%;
+    height: auto;
+    bottom: 0;
+    left: 0;
+}
+
+.sm-hero__grid__grid-item--intro h1 {
+    padding-left: 80px;
```
LL782, line 435: I notice that this style gets overridden further down this same file in the `(min-width: 992px)` media query.

Why not move it to the `max-width: 991px` media query?

src/assets/custom/css/custom-softmod-feature.scss
```diff
@@ -0,0 +1,751 @@
...
+    .sm-hero__grid__grid-item--intro a {
+        margin-left: 6rem;
+    }
+}
+
+@media (min-width: 992px) {
+
+    .sm-hero {
+        background-color: transparent;
+        background-image: none;
```
LL782, line 456: These two rules are clearly an over-ride. Why not move the rules that are being overridden into the `max-width: 991px` breakpoint where they belong?

LL782 (commented):
@keith-smale I've only made it half way though so figure I'll share my comments with you so far. There's nothing blocking that I've found but there is a reference to an asset that seems to be missing or unneeded.

src/_includes/head.html
keith-smale: Yes the meta on the site needs a complete review and this is by no means how I would like it to be. But to get this out I made these changes for now. I would want the logic to calculate the right text for the title's, descriptions etc. to sit in a separate file and just pull in the result into a variable for each meta tag value.
The page title tag could be all on one line (which it was to start with) but I left with a separate part of the if on each line just for readability of the if statements.

src/_includes/head.html
keith-smale: Yeah ideally we should just have a variable for the result of the if, injected into the meta.

src/_includes/head.html
keith-smale: Noted an as per the previous comments. Also happy to take any suggestions on how it should be implemented to support the English and Spanish versions of text plus any default values.

src/_includes/header.html
keith-smale: It is yes, but we can obviously alter the naming to be more clearer

src/_includes/head.html
anguspaterson: It's very convoluted. We are battling against two boilerplate unity themes. The whole thing needs to be inspected and potentially rebuilt from the ground up

src/assets/custom/css/custom-softmod-feature.scss
anguspaterson: I wrote a helper class to address this, it should be used instead

src/assets/custom/css/custom-softmod-feature.scss
anguspaterson: These all come with the theme we are using. I've tried to use existing classes as much as possible and not introduce new ones and then follow the naming convention that is already set out. Not sure what g means but bp is padding bottom.

src/assets/custom/css/custom-softmod-feature.scss
anguspaterson: The text changes colour on the breakpoint - appreciate this needs to be rewritten/refactored and the !important removed

src/_includes/head.html
LL782: Cool. We're in agreement

src/_includes/head.html
LL782: Yep. We definitely need to look at how to resolve these issues. I left my comments to try to get that conversation started. @keith-smale suggested that perhaps now isn't the time to take this on – it should be a project in it's own right. Hopefully next week I'll have some time to start making suggestions about how we do that. 

I'd prefer an iterative approach to a clean break if we can come up with a straight forward strategy for doing it.

PR #1706, opened 2020-04-24
Softmod publications tag injestion

mattgraygithub:
This adds functionality for pulling in tagged posts into the software modernisation page and puts this behind the feature [toggle](https://github.com/codurance/site/blob/master/src/_data/toggles.yml). Currently the tag is set to "life at codurance", but when "software modernisation" tag exists, it should be changed to this. Note that the feature toggle is currently off.


LL782 (approved):
Looks good to me! I got a walk through from @keith-smale and have had it running locally behind the feature toggle. 

In future I think we should ask @anguspaterson what the designers are calling the `tagged_posts` module and probably rename that file but let's not delay integration for that. We can make those changes easily later 

Let's get this merged!

🍕

keith-smale:
Looks good to me.

keith-smale (approved):

PR #1709, opened 2020-04-27
Refactor metadata content variables

keith-smale:
This is a refactor to how the metadata is managed on the site. The meta tags in the head.html page had a lot of logic in them and this got worse when we made changes for the Software Modernisation page.

This change removes the logic out of the head.html page and just fills liquid variables for most of the tags.
The liquid logic is moved to a new meta-data.html file and has been simplified to make it cleaner and easier to understand.

This change paves the way for a site wide update of metadata to improve the sites SEO.


src/_includes/meta_data.html
```diff
@@ -0,0 +1,69 @@
+<!------Start Title Meta------->
```
LL782: Regarding these comments:
- If they're needed (not sure they are) perhaps we should have multiple includes and turn the comments into the file names, e.g. `meta_titles.html`, `meta_description.html`, `meta_share_images.html`
- Some of the comments are not considered valid XML. The character sequence '--' shouldn't be used inside a comment. This seems to upset some browsers (see image) _This was news to me but looks funky if anyone inspects our source code_
- If anyone does inspect out source code these comments will appear totally out of context as they rap variable definitions that will not appear (see image). This is very minor but I guess some of those customers will be scrutanising our code. It's also unnecessary extra weight in the download.

Screenshot 2020-04-27 at 14 36 30

LL782 (approved):
@keith-smale so happy to see this change.

I've left a comment about comments. Let's get this in but perhaps address that on a new branch. Or I can do it if you support the suggestion of splitting them into multiple includes.

src/_includes/meta_data.html
LL782: Resolved by https://github.com/codurance/site/pull/1709/commits/8abbbe522261704d528ac5ec386aa50bdf1d36cf

LL782 (approved):

PR #1711, opened 2020-04-27
Avoid toggles being changed from local development

LL782:
To:
- make it harder to accidentally change a feature toggle that affects prod and 
- make it safer to change a feature toggle locally

This PR also contains addtions to the README

@keith-smale @Dan-Bird
Following approval and merge of this PR we should create a fresh `src/_data/toggles.yml` via GitHub with the following contents:

```yml
feature-sm-blogs: "off"
```

This is not necessary but would be helpfully, ready for when we need to turn the first feature on.

_Reviews from all Codurance welcome!_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,5 @@ tools/snapshots/backstop_data/*
 output/*
 
 coverage/*
+
+toggles.yml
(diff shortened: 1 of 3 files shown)


LL782:
Closing in favour of https://github.com/codurance/site/pull/1712

PR #1714, opened 2020-04-28
Add Feature Toggles notes to README

LL782:
This just adds a note to the README about the current use of Feature Toggles

Preview here:
https://github.com/codurance/site/blob/feature-toggles-readme/README.md#feature-toggles


anguspaterson:
Look good to me and makes sense in terms of instruction

PR #1716, opened 2020-04-28
Start adding data for the Softmod Publications block

LL782:
Four real changes here:

- **Apply the name "Curated Publications" to the UI component previously "Tagged Posts"**
  This comes from a discussion with @anguspaterson regarding how this component is intended to be used (not necessarily coupled to tags)
- **Use a "Software Modernisation Curated Publications" component on the Software Modernisation page**
  This passes in the `tag`, `title` and `description` specific to Software Modernisation 
- **Add keys for the i18n text needed by this module**
  This text still needs to be translated but now we have the keys and references to them inside the component
- **Use the new "Learn More" translation `translation publications.teaser-cta` for the existing publication cards on the homepage**
  Previously this was hardcoded in English only

Please can I get one or two approvals on this before merging


Dan-Bird:
This looks good to me

PR #1717, opened 2020-04-28
Update clients page buttons and links

mattgraygithub:
This satisfies tickets 106, 107 and 108, 109 and 110 of the software modernisation leankit board. 

106 - remove existing case study pages for ASOS
107 - Redirect the case study button for ASOS to new case study page (it was later agreed to simply delete the old ASOS case study pages and change the button link, rather than redirecting the old link to the new one)
108 - Add a button on the Mango and ECW section which leads to each case study
109 - Add new block for bestsecrets which leads to it's case study page
110 - Update call out phrases for each client and add software modernisation


keith-smale:
This looks good to me.

keith-smale (approved):

PR #1719, opened 2020-04-28
Class names and markup for first iteration of software modernisation curated publications

LL782:
Done:
 - create the structure for the SCSS and defines mark-up for the teaser cards
 - creates a main.css file into which we can import scss partials
 - removes the .card rules from bootstrap.css because I was able to confirm we're not using them


anguspaterson:
Discussed with the author. This is good to go and makes great sense.

anguspaterson (approved):
Looks great and have discussed with author! Makes great sense

PR #1732, opened 2020-05-03
Updated profile pic and description

ghost:
as per title


sherlockq:
Though I'm not sure we have a written rule on the aspect ratio of image, it may be better to be consistent. Keep the parrot and see how much we left with the glass!

Mattsi-Jansky:
Sherlock is right, images should be 400x400. Otherwise it breaks like so
Screenshot 2020-05-04 at 08 44 30

src/_data/team.yml
```diff
@@ -734,7 +734,7 @@
     <p>Sylvester started programming by learning Turbo Pascal in the late 90's. He moved to Borland Delphi 5 in the early 2000s where he programmed (for fun) two games, Snake (as per Nokia) and Space Invaders. Around this time, Sylvester was exposed to LAMP and the .Net suite where he learned Visual Basic, Visual C++ (Direct X), C# and ASP.Net.</p>
     <p>Since 2012, Sylvester has been developing Web and Ops Enterprise solutions in multi-disciplined teams, working closely with the business.</p>
     <p>Sylvester has engaged across the entire product lifecycle from defining the requirements with Product Owners all the way to maintaining them in production.</p>
-    <p>In his free time, Sylvester enjoys learning and sharing knowledge around multiple subjects including martial arts and how to be a better version of himself.</p>
```
Mattsi-Jansky, line 737: "and how to be a better" has one "a" too many. Perhaps "and self-improvement"?

Mattsi-Jansky (changes requested):
Dig the new picture. Asked for a couple small fixes.

ghost:
@Mattsi-Jansky @sherlockq 
As per your recommendation, I updated the image ratio accordingly.
Thank you for your feedback!

Screenshot 2020-05-05 at 18 40 25

Mattsi-Jansky (approved):
Think you left an errant `"` character in that last change, but otherwise it looks good 👍

PR #1736, opened 2020-05-06
show article title and description in es on curated publications

keith-smale:
Added logic to show article titles and descriptions on the curated publications block (in the soft mod landing page) in the correct language based on if English and Spanish language is selected.

Need to update the logic to cater for a default scenario in case there is no Spanish title or description for the article but the site is viewed in Spanish, but this will work to get the article block live.


LL782 (approved):
Looks good!

Is camel case the standard casing for Jekyl/Liquid?

PR #1738, opened 2020-05-06
Correct definition of View All Publications href

LL782:
This fixes a defect where we were not correctly setting the href for the link at the bottom of the Curated Publications module.

The fix uses Liquid's [`capture`](https://shopify.dev/docs/themes/liquid/reference/tags/variable-tags#capture) tag for better string interpolation.

@jpwenzelc @Dan-Bird


Dan-Bird (approved):
This all looks good to me 👍

PR #1741, opened 2020-05-06
Soft mod/refine spacing

LL782:
Further refinement of publication cards to allow for long titles:

- Introduce a consistent max-widths to Soft Mod page
- Provide three lines for card titles on large screens
- Only fit three cards on extra large screens (rather than four)

@anguspaterson please review the adjusted designs

---

**Large screens** with three lines of space for the title (the cards are quite narrow on this screen size):

Screenshot 2020-05-06 at 19 06 39

**Extra large screens** only show three cards to prevent cards from being too narrow:

Screenshot 2020-05-06 at 19 06 14

---

We've also changes the Spanish text for the CTA on the Publication Cards

### DIFF
--- a/src/assets/custom/css/_sass/_card.scss
+++ b/src/assets/custom/css/_sass/_card.scss
@@ -40,7 +40,7 @@ $line-height: 28px;
   }
 }
 
-$description-max-height: 7 * $line-height;
+$description-max-height: 4 * $line-height;
 
 .card__description-fade-out {
   overflow: hidden;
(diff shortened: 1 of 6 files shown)


src/assets/custom/css/_sass/_card.scss
```diff
@@ -22,14 +22,25 @@
   flex-grow: 1;
 }
 
+$line-height: 28px;
+
 .card__title {
+  $line-height: 28px;
```
LL782: We don't need this defined twice, I'll remove it

anguspaterson (approved):
This looks good to me and is as the desired outcome as discussed

PR #1744, opened 2020-05-11
Develop Webinar Hero Teaser

Dan-Bird:
This sets up the base for the Webinar Hero Teaser. Further styling is needed.


LL782 (approved):
Let's get this in! :pizza:

PR #1749, opened 2020-05-12
Webinar Teaser - Layout refinements

LL782:
These two changes come out of a design review:

1. Use the same "sheen" background behind both sections, to avoid a "hard edge" where the gradient repeats itself
1. Apply the same, "pull-up" effect to the Webinar Teaser as we had with the Curated Publications section, so the form slightly overlaps this section.

See image for reference

Screenshot 2020-05-12 at 12 43 06

Dan (and anyone else) please review and ask about anything you're not sure off.

**Note** you probably want to _Hide Whitespace Changes_ when you review to make the changes in `soft_mod_webinar_hero_teaser.html` easier to read.


Dan-Bird (approved):
These changes all seem good to me.

PR #1750, opened 2020-05-12
Soft Mod Webinar Teaser – Add translations and link, and release feature

LL782:
The title says it all really. This is the last change needed before we can release the feature.

@keith-smale @nicholemellekas @Dan-Bird @anguspaterson 
This is behind a feature toggle so I reckon we go to production with it and do any final reviews there.


Dan-Bird (approved):
Looks fine to me 👍

PR #1751, opened 2020-05-12
Publications header tweaks

anguspaterson:
- Separate date and meta from title/lead and house in separate container
- Add calendar icon to date
- Add overlay to the banner to increase text legibility of H1 / Lead
- Remove min-height and replace with a more flexible padding solution


keith-smale (approved):
Looks ok. Better than what we had.

PR #1752, opened 2020-05-12
Nichole post 5 signs

mattgraygithub:
Adds the blog post "5 signs that it is time to launch a modernisation strategy" by @nicholemellekas


keith-smale:
Looks lovely

PR #1755, opened 2020-05-12
Soft mod/pin articles

keith-smale:
Updated the way that the curated publications block on the software modernisation page is populated. The change swaps out looking for content with the 'software-modernisation' tag to use a pinned_locations list added to a post.

Articles with a pinned_locations list item of 'soft-mod' articles are shown on the software modernisation curated publications block and will remain in that block until the 'soft-mod' list item is removed from the post front matter section.


LL782 (approved):
🎱 LGTI

PR #1760, opened 2020-05-13
Add drop shadow to Hero Teaser

Dan-Bird:
**What**
This PR adds a drop shadow to the white card sections of the Hero Teaser component upwards of large screens. 

This PR also unifies the shadow that's used within the cards holding the Curated Publications and the content of the Hero Teaser as mentioned above. 

**Why**
This is to improve contrast between the white of the card and the background it sits on. On less quality monitors and displays the contrast between these two elements may not be noticeable.

Screenshot 2020-05-13 at 17 59 46

### DIFF
--- a/src/assets/custom/css/_sass/_card.scss
+++ b/src/assets/custom/css/_sass/_card.scss
@@ -1,7 +1,11 @@
+@mixin style-card-box-shadow {
+  box-shadow: rgba(0, 0, 0, 0.16) 0 0 8px;
+}
+
 .card {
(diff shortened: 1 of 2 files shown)


LL782 (approved):
Looks good! Let's get this in. 

Minor notes for the future: 

I'm not sure about the name `style-card-box-shadow` so let's keep an eye on this. Also agree with what you said on Slack @Dan-Bird about moving this mixin to a partial. Currently it's quite a fragile link between _card and _hero-teaser

PR #1764, opened 2020-05-14
Remove unecessary target blank

LL782:
These changes fix an issue noticed by @mattgraygithub where the **Hero Teaser** of the **Curated Publications** always opens in a new tab.

I've also updated the README since the recent Javascript changes that add target _blank to all external links.


Dan-Bird (approved):
All good to me

PR #1765, opened 2020-05-15
Apply Section Intros UI in place of bespoke HTML and CSS

LL782:
In this PR we extract the HTML for the Section Intro into it's own file and replace some existing HTML and CSS. 

There was also some quick clean up of the Testimonials element on the Soft Mod page introducing the shared breakpoints and CSS approach.


Dan-Bird (approved):
All good with me

PR #1766, opened 2020-05-15
Cheat sheet cta banner

Dan-Bird:
This PR duplicates the CTA Banner currently being used within the Soft-Mod campaign, to create a version for Business Case Questionnaire / Cheat Sheet download. 

This can now be included similarly to the existing CTA Banner.

Since there is duplication of components - next steps will be to extract this into a re-usable component, and replace existing use of this component.


LL782 (commented):
I'm confused because this component doesn't seem to be being used. i.e. Where is it included?

src/_includes/soft_mod_cheat_sheet_cta_banner.html
```diff
@@ -1,6 +1,6 @@
 <div class="container section g-pt-200 g-pb-100">
   <div class="soft-mod-in-page-banner">
-    <div class="soft-mod-cheat-sheet-cta-block-image"></div>
+    <div class="soft-mod-cheat-sheet-cta-block-image-{{ site.baseurl | remove: '/' }}"></div>
```
LL782: Does this work in production? I think `site.baseurl` is empty outside of local development

LL782: If it doesn't work you might be able to use `site.lang` but it would be good to have this code wired into the website someone to be able to check

src/_includes/soft_mod_cheat_sheet_cta_banner.html
```diff
@@ -0,0 +1,30 @@
+<div class="container section g-pt-200 g-pb-100">
```
LL782: It would be nice to use a single BEM style block classname for these styles.

As the file is called `soft_mod_cheat_sheet_cta_banner` that would seem to be the block name. Is that a name that @anguspaterson is using in the designs? Happy to look at it together sometime

Dan-Bird:
It will be used within some of the Blog Posts for soft mod. It's not currently included in one. We will look to include it in José's article.

Maybe I can create a toggle for it to be visible.

src/_includes/soft_mod_cheat_sheet_cta_banner.html
```diff
@@ -0,0 +1,30 @@
...
+  <div class="soft-mod-in-page-banner">
+    <div class="soft-mod-cheat-sheet-cta-block-image-{{ site.baseurl | remove: '/' }}"></div>
+    <div class="soft-mod-cta-block__main">
```
LL782: This classname suggests it is the `__main` element of a BEM style block called `soft-mod-cta-block`. Where is that block?

src/assets/custom/css/_sass/_soft-mod-cheat-sheet-cta-banner.scss
```diff
@@ -0,0 +1,36 @@
+.soft-mod-cheat-sheet-cta-block-image-en {
```
LL782: The classnames in this file seem inconsistent. The file name, the related HTML file and the different classnames, I'd like to see these using the kind of consistency we planned for `section-intro`

src/assets/custom/css/_sass/_soft-mod-cheat-sheet-cta-banner.scss
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  @include medium-large-and-extra-large {
+    background-image: url("#{$site-base-url}/assets/custom/img/softmod/bg-cta-banneer-business-case-for-change-es.jpg");
+    background-position: top left;
+    background-size: cover;
+    float: left;
+    height: 400px;
+    width: 50%;
+  }
```
LL782: The only thing that varies between the `en` and `es` versions is the image. The rest of the styles could be shared

Dan-Bird:
I've included it within the most recent blog post so it can be previewed

LL782 (approved):
I've left a bunch of comments. I'd be happy to go through these in person. I'm also fine if you merge it and we improve this in future.

LL782 (approved):
🍕 Nice changes!

PR #1768, opened 2020-05-18
Quick UI design consistency improvements

LL782:
Two changes made while pairing with @anguspaterson 

1. Bring the max-width used in the latest designs in line with the existing max-width controlling there header (at this point we've only addressed this for large and extra-large screens
1. Minor adjustment to the line-height of the `mercer` type style


anguspaterson:
This is good to go. Looks good and as discussed to bring in line current grid / new design

anguspaterson (approved):

PR #1775, opened 2020-05-21
update compass cta on soft mod page to use compass-es if site is in S…

keith-smale:
…panish

This change adds a condition to direct users to the Spanish version of Compass if the site is being viewed in Spanish.


Dan-Bird (approved):
All good with me 👍

PR #1776, opened 2020-05-21
small change to update the compass URL for the ES version.

keith-smale:
Request to update the compass URL for the Spanish version: 
From using `compass-es.codurance.com`
To use `compass.codurance.es`


Dan-Bird (approved):
All good :)

PR #1778, opened 2020-05-21
Fix: Accessing sections of the page specified by the hash within the URL

Dan-Bird:
This PR adds a new component, the Page Location Anchor. 

This is used to add an on-page anchor to specific sections that we would like the browser to take the user to, or if we would like to create links directly to the specific section.

It takes the site's navigation bar into account so that it will no longer overlap the content we wish to be displayed.

### DIFF
--- a/src/_includes/compass_teaser.html
+++ b/src/_includes/compass_teaser.html
@@ -1,4 +1,7 @@
-<section id="compass-by-codurance" class="compass-teaser">
+<section class="compass-teaser">
+  {%- assign anchor_id = "compass-by-codurance" -%}
+  {% include page_location_anchor.html anchor_id=anchor_id %}
+
   <p class="compass-teaser__description">{% t software-modernisation.compassCTA-subheading %}</p>
   {% if site.lang == 'es' %}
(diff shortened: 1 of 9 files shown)


src/_includes/compass_teaser.html
```diff
@@ -1,4 +1,7 @@
-<section id="compass-by-codurance" class="compass-teaser">
+<section class="compass-teaser">
+  {%- assign anchor_id = "compass-by-codurance" -%}
```
keith-smale, line 2: What is the use of the hyphen after the and before the %?

I've seen that for all the assign statements  for the anchor_id but I wasn't sure why they were needed.

Dan-Bird: I added Jekyll Snippets to my VS Code, and they were put in automatically. 

I also wondered why they were being put in and when I looked it up it mentioned that it removes any generated white-space from the template.

keith-smale: The scrolling works so it's not causing an issue in the code but just different to the documentation for liquid.

keith-smale (approved):

PR #1780, opened 2020-05-22
[Merge on 26th of May] Blog Entry "Hidden Cost of Not Modernising"

josewenzel:
The translation is in the metadata but is not yet implemented, so once that's done it will apply the Spanish `title` and `description` to the Spanish version of the site.


LL782 (approved):
I haven't reviewed the article content the markdown and front matter all look sane

PR #1783, opened 2020-05-26
Use placeholders for tabber content

LL782:
These changes are part of replacing the Soft Mod content with placeholders as per this ticket: https://codurance-online.leankit.com/card/1091951812

Some comments were also deleted in the process


Dan-Bird (approved):
Looks fine 👍

PR #1784, opened 2020-05-26
Remove Webinar and Publications from Service Line template

LL782:
These changes adjust the CSS so that the overlapping layout is correct with or without the Webninar and Curated Publications modules.

Also on this branch we've removed those modules from the example template:
https://codurance.com/services/service-line-template


src/assets/custom/css/_sass/_soft-mod.scss
```diff
@@ -84,7 +89,24 @@
...
+
+  @include small {
+    padding-top: 50px;
+    padding-left: 15px;
+    padding-right: 15px;
+  }
+
+  @include medium-large-and-extra-large {
+    margin-top: -150px;
+    padding-top: 100px;
```
LL782: We could use the overlap variable to explain these two values.
```suggestion
    margin-top: #{(-1.5 * $vertical-overlap)};
    padding-top: $vertical-overlap;
```

Dan-Bird (approved):
All looks good

PR #1786, opened 2020-05-28
Update Services page form

LL782:
These changes relate to [this ticket on Lean Kit](https://codurance-online.leankit.com/card/1093152584) 
**"Add Compass Teaser and latest contact form design to Service Page"**

Also done:

- The Contact page uses the same HubSpot form ID as before
- The Service page now uses the same HubSpot form as the Software Modernisation page
- The UI for both of these forms has been brought inline with the latest designs
- The Contact page layout has been slightly adjusted to fit better with the new designs
- The Compass Teaser fits in between the contact form and the rest of the Services page
- Some dead code deleted


src/_includes/get_in_touch.html
```diff
@@ -0,0 +1,51 @@
+<section class="get-in-touch">
+  {% include page_location_anchor.html anchor_id="get-in-touch" %}
+
+  <div class="get-in-touch__form">
+    {% include section_intro.html description=include.description title=include.title %}
+
+    <div class="sm-enquiry">
```
LL782: Not a blocker but it would be nice to bring this class name in line without our BEM style block,
e.g. `get-in-touch__form-inner` or something better

Dan-Bird (approved):
All looks good 👍

PR #1789, opened 2020-05-29
Move client section down page

Dan-Bird:
This PR prepares the Service Offering page to be worked on by first moving the Client section down the page.


josewenzel (approved):
All good

PR #1791, opened 2020-05-29
Update client section service offering

Dan-Bird:
This PR updates the client section of the Service Offering page. 

It also updates the Section Intro component so that it can be reused within other components with an optional description.


LL782 (approved):
Code looks great! I haven't checked out the branch and tested in the browser. I wonder about how the change from sheen to slate looks but I know you guys paired on this so if you're happy I am

PR #1793, opened 2020-06-01
Service Line Hero Teasers enhancements

LL782:
With this changes we split the **Teaser Group** (new component) from the **Hero Teaser** (existing component that we were modifying)

At the stage of this PR the Teaser Group is near enough a straight copy of the Hero Teaser with two major changes:

1. It has the slate colour background
1. It expects and icon to be passed into the include

After this changes, the Hero Teaser is not expected to be modified any further for the new Service page design. Further changes will be made to the Teaser Group.

Suggestion for future work:

- Rather than using four teaser group components to make up the service line layout, use a single teaser group and pass data into it for the four service lines


Dan-Bird (approved):
All good to me

PR #1794, opened 2020-06-02
Update hash links for Soft-Mod

Dan-Bird:
This PR puts the IDs for the Page Location Anchor component within the site's translation files. 
This way they can be customised between the EN and the ES version of the website.


LL782 (approved):
:+1:

PR #1800, opened 2020-06-04
Add bundle install error solution to  README.md

josewenzel:
There are some people having trouble with `rake servequick`, we added the solution to the `README.md` 💪


LL782 (approved):
Looks sane

PR #1808, opened 2020-06-05
First draft services content

LL782:
This puts the latest images, icons and text content into the Services pages. Note most of this is behind the feature toggle. 

With these changes we've simplified the SVG icons to.


Dan-Bird (approved):
# 👍

PR #1809, opened 2020-06-08
Update page intro on Service page

LL782:
This brings the new Page Intro UI component to life on the Service page.

With these changes we have also refinded the "out-there" max width components to ensure they all line up correctly and have consistent spacing.

Notes:
- This is all behind the `service-line-updates` feature toggle
- Text and images are used for example purposes only
- The was one area of the design that wasn't clear (at 1300px wide). We should review with @anguspaterson


Dan-Bird (approved):
This is all good with me 👍

PR #1815, opened 2020-06-10
Fix small errors

Dan-Bird:
This PR fixes two small errors within the latest blog post

- hyperlink not displaying correctly
- mentioning an image below rather than above


LL782 (approved):
:+1:

PR #1816, opened 2020-06-10
Include Business Case Presentation In-Page Banner

Dan-Bird:
This PR modifies the In-Page Banner component so that it can be used within the Business Case blog post to promote the Business Case Presentation Template download. 

The component itself can now take an optional tag within the include to specify whether to display in the new right-heavy split layout. By default it will display as the evenly split layout it had before. 

The include within the blog post still requires the correct link to the Business Case Presentation Template document within Hubspot.


Dan-Bird:
This branch has been pulled into newer branches to take advantage of the modifications of the In Page Banner. 

This branch will either need to be updated - or a new one created which just includes the In Page Banner Include from commit [17b3f8c](https://github.com/codurance/site/pull/1816/commits/17b3f8cf992e13b203e89da7eb193e9bb627c3ac)

LL782 (approved):
I haven't previewed the branch but code all loks sane

PR #1818, opened 2020-06-11
Create service line layout

LL782:
These changes introduce a layout for the Service Line pages and begin the process of reducing the repetition between the "service-line-template" example and the live "software-modernisation" page.

We're intending to come back for further work on the sheen section but now as part of this PR


Dan-Bird (approved):
# 🕵🏼

Looks good to me

mattgraygithub (approved):
LGTM 👍 👍

PR #1821, opened 2020-06-12
Tidy service line template

mattgraygithub:
This is to better organise our styles by renaming the SCSS classes used in the software modernisation page and service-line-template page to be generic for all service lines and not just software modernisation.


Dan-Bird (approved):
Good with me - Further refinement and improvements can be done along the way visiting each section we include within the Service Line Template

PR #1826, opened 2020-06-15
Update copy on services page

Dan-Bird:
This PR adds the copy for the Services page and the Services teaser group content.


LL782 (approved):
Looks sane :+1:

PR #1827, opened 2020-06-15
Add updated imagery to page

anguspaterson:
Updated imagery is now included to the page for the hero section and subsequent service line hero slices


LL782 (approved):
Looks good!

PR #1831, opened 2020-06-15
Service page social images

LL782:
These changes add the social share image, the `og_image`, for the Services page.

As this is the first time we have had locale text in the image, these changes also introduce the logic to vary the `og_image` based on the language of the website.

The `twitter_image` has also been sustituted in favour of the higher resolution `og_image`. This is clean up of a minor defect that we've been living with.


Dan-Bird (approved):
# 🏎️ 

All good with me

PR #1833, opened 2020-06-16
Add Hero Section to the Service Line Template

Dan-Bird:
This PR adds the Hero / Intro section to the Service Line Template. 

The section is now within it's own include. The styles have been refactored to use BEM naming convention to better target elements within the CSS. The styles have also been cleaned up so that they now follow our new breakpoints and typography styles - keeping everything consistant throughout the site. 

It picks up information about which images to use from the front matter. Setting a service-line property on the front matter of an intended service line page will pull in the relevant icon for it, and gather the copy from the translation files.


LL782 (approved):
Nice :+1:

mattgraygithub (approved):
Looks good! 💥

PR #1835, opened 2020-06-16
Remove broken redirect

LL782:
We found that our redirect of `/services/service-line-template` is not working as expected. Rather than burn time looking into it we're simply removing it to not cause confusion in future.


mattgraygithub (commented):
👍

PR #1837, opened 2020-06-17
Remove "service-line-updates" feature toggle

LL782:
Context available via: https://codurance-online.leankit.com/card/1111483649

With these changes we delete code that is no longer needed and remove the feature toggle, so the new feature has fully substituted the previous version.

Notes:
- We removed CSS, HTML and images that are no longer used
- We also removed the "We advise, we coach, we deliver" block from the homepage
- We changes the angle of the `$sheen` background gradient so that blocks with this background can be stacked one on top of the other
- We generally cleaned up and introduced consistency to the Services page and components related to the Compass Teaser
- We also deleted `.services-hero-title` CSS. This was used in a number of inappropriate places, so we've replaced it with `.g-color-white` as these components are already using loads of Bootstrap class names already


Dan-Bird (approved):
All good

mattgraygithub (approved):
Looks good to me 🌮

PR #1838, opened 2020-06-17
Update cookie name to match cookie that is used

cherryt:
The cookie that is created when a user accepts cookies is called 'has-cookie-consent' not 'seen-cookie-message'. Updated the name in the Cookie Policy page to reflect that.


LL782 (approved):
Looks sane!

mattgraygithub (approved):
Looks good to me 👍

PR #1842, opened 2020-06-17
Service line template/put data in yml

Dan-Bird:
This PR relates to the following [corresponding ticket](https://codurance-online.leankit.com/card/1123495562).

It creates a data file specific for the tabber component that will be used within the Service Line Page Template - and therefore throughout the Service Line pages. 

The tabber component within the template will look at the front matter within the service page using it, and use the `service-line` property there to pull in the relevant information from the data file.


src/_layouts/service_line.liquid
```diff
@@ -3,6 +3,8 @@ layout: default
 ---
 
 {% assign translations = site.translations[site.lang].[page.service-line] %}
+{%- assign all-service-lines = site.translations[site.lang].all-service-lines -%}
```
LL782: It's not clear to me why we need to assign `all-service-lines` here (or at all really)

src/_i18n/en.yml
```diff
@@ -775,3 +713,13 @@ service-line-example:
...
+
+all-service-lines:
+  page-sections:
+    how-we-solve-your-problems: how-we-solve-your-problems
+    how-we-help-our-clients: how-we-help-our-clients
+    testimonials: testimonials
+    compass: compass-by-codurance
+    webinar: webinar
+    publications: publications
+    get-in-touch: get-in-touch
```
LL782: Looks like these need to be updated in the HTML as well. Missing commit perhaps?

LL782 (commented):
I'm only part way through the review. It's worth me sending what I've got so far...

src/_data/service_lines_tabber.yml
```diff
@@ -0,0 +1,125 @@
...
+    tab-six-intro: Risk management, operational excellence, security, and compliance must be done by design, not as an afterthought
+    tab-six-list-item-one: Real-time system intelligence
+    tab-six-list-item-two: System observability
+    tab-six-list-item-three: Security
+    tab-six-list-item-four: Operational excellence
+    tab-six-list-item-five: Reduce mean time to recover
+    tab-six-details: Isolation and tighter control of critical areas of the system; Strategic approach to regulatory compliance, security, monitoring, auditing and logging; Operational effectiveness and cost control; Site Reliability Engineering (SRE)
+    tab-six-image-path: /assets/custom/img/softmod/bg-risk-management.svg
+
+es:
```
LL782, line 64: Really like the decision to separate by language :+1:

src/_i18n/en.yml
Dan-Bird: I've updated the HTML to now reference these.

src/_includes/compass_teaser.html
```diff
@@ -1,6 +1,6 @@
 <section class="{{ include.class }}">
   <div class="compass-teaser">
-    {%- assign anchor_id = {site.translations[site.lang].software-modernisation.page-sections.compass} -%}
+    {%- assign anchor_id = all-service-lines.page-sections.compass -%}
```
mattgraygithub: How does it get the translations from all-service-lines? Do we need a `{t all-service-lines.page-sections.compass} ` for this?

src/_layouts/service_line.liquid
mattgraygithub: Ah does this answer my question above?

src/_data/service_lines_tabber.yml
```diff
@@ -0,0 +1,125 @@
...
+  how: How
+
+  software-modernisation:
+    tab-one-title: Sustainable Change
+    tab-one-intro: Business can move only as fast as we can deliver quality software.
+    tab-one-list-item-one: Reduce lead time
+    tab-one-list-item-two: Reduce cost of change
+    tab-one-list-item-three: Frequent and safer releases to production
+    tab-one-list-item-four: Reduce technical debt
+    tab-one-list-item-five: Reduce defect rate
```
LL782, line 12: Either now or part of the clean-up task, what do you guys think to doing something like this?

@mattgraygithub @Dan-Bird 

```suggestion
    tab-names:
      - Reduce lead time
      - Reduce cost of change
      - Frequent and safer releases to production
      - Reduce technical debt
      - Reduce defect rate
```

I'm not 💯 on the syntax but using a YML list It would gives us more flexibility.
For example, `tab-one` here has a longer list than `tab-two`.

src/_includes/compass_teaser.html
mattgraygithub: Maybe to make it clearer we could not assign `all-service-lines` in the `service_line.liquid` layout and instead do something like `anchor_id = translations.common-service-line-content` ?

src/services/software-modernisation/index.html
```diff
@@ -20,51 +20,73 @@
 
 service-line: software-modernisation
 ---
+{% assign translations = site.translations[site.lang].[page.service-line] %}
+{%- assign all-service-lines = site.translations[site.lang].all-service-lines -%}
```
mattgraygithub: Now a bit confused whether this needs to be in service_line layout as well as here

src/_data/service_lines_tabber.yml
mattgraygithub: Just a question and not a suggestion - why don't we put these in the translations?

mattgraygithub (commented):
Just have a few questions I'd like to talk about

src/_includes/compass_teaser.html
LL782: Where is `all-service-lines` defined? If we have a Compass Teaser outside of the Service Line pages will this still work/make sense?

LL782 (commented):
Couple more comments. Aside from these it looks all good!

src/_data/service_lines_tabber.yml
LL782: The `en` and `es` files are getting massive and are hard to read. Keeping this stuff encapsulated and "close" to where it's used makes it easier to maintain. I say "close" because it has the same name as the component where it's used.

src/_includes/compass_teaser.html
LL782: Personally I prefer something more simple, like 
```suggestion
    {%- assign anchor_id =  site.translations[site.lang].compass-teaser.anchor-id -%}
```

LL782:
@Dan-Bird looks like there are conflicts to be resolved ☹️

src/_layouts/service_line.liquid
Dan-Bird: I have assigned it both here and within the index page for Software Modernisation, as the index page doesn't pick it up from the Service Line Template. 

I was thinking that when we move the markup from the index page into the template, then we can remove the assignment within the index page. 

I can remove this assignment here and we can just swap out where the assignment is made when we come to moving the markup from the softmod index page into the Service Line Template.

src/_includes/compass_teaser.html
Dan-Bird: This has been actioned now. The Compass Teaser component now pulls in it's anchor independently from the service lines. 

I've also changed it to pull in it's text from a compass-teaser entry within the translation files as it's independent of the service lines at this point too.

src/services/software-modernisation/index.html
Dan-Bird: This assignment has been removed, however these types of assignments do need to be assigned in both the template and softmod page as there is content on both that needs it to be defined in those sections until we move the markup to the template. 

Happy to demonstrate further for clarity. 

It seems that anything pulled in through `{{ content }}` cannot pick up any assignments made in the layout, where as includes can.

src/_data/service_lines_tabber.yml
Dan-Bird: I'd be keen to speak about the intentions here. I'm not quite sure I understand the use. 

This could be done as part of the cleanup though.

src/_layouts/service_line.liquid
Dan-Bird: Please see my other comment referencing assignments not being picked up by anything being pulled in through `{{ content }}`

src/_includes/get_in_touch.html
```diff
@@ -1,4 +1,4 @@
-{%- assign anchor_id = {site.translations[site.lang].software-modernisation.page-sections.get-in-touch} -%}
+{%- assign anchor_id = all-service-lines.page-sections.get-in-touch -%}
```
LL782, line 1: Should we move this into a `get_in_touch` part of the i18n files?

src/_includes/service_line_details_section.liquid
```diff
@@ -1,5 +1,11 @@
+{%- if site.lang == 'es' -%}
+  {%- assign tabber-data = site.data.service_lines_tabber.es.[page.service-line] -%}
+{%- else -%}
+  {%- assign tabber-data = site.data.service_lines_tabber.en.[page.service-line] -%}
+{%- endif -%}
+
 <section class="service-line-problems">
-    {%- assign anchor_id = {site.translations[site.lang].software-modernisation.page-sections.how-we-solve-your-problems} -%}
+    {%- assign anchor_id = service-line-page-locations.how-we-solve-your-problems -%}
```
LL782, line 8: Where does `service-line-page-locations` come from?

src/_layouts/service_line.liquid
```diff
@@ -3,6 +3,13 @@ layout: default
 ---
 
 {% assign translations = site.translations[site.lang].[page.service-line] %}
+
+{%- if site.lang == 'es' -%}
+  {%- assign service-line-page-locations = site.data.service_line_page_section_anchors.es -%}
+{%- else -%}
+  {%- assign service-line-page-locations = site.data.service_line_page_section_anchors.en -%}
+{%- endif -%}
```
LL782, line 11: This is causing me to scratch my head a bit

LL782 (approved):
All works and loads of good work we want to get into master, so let's merge this and come back for the comments now on a new branch?

src/_includes/get_in_touch.html
LL782: We resolved this as part of 6ecbd53e6f07c167ece994caef64bc8805549a5a

PR #1844, opened 2020-06-22
Move Case Studies section to Service Line Template

Dan-Bird:
This PR moves the case studies section of the service line pages into the Service Line Page Template.

It also moves the translations into a section that is not specific to Software Modernisation since it will apply across all service Lines.

The styling has been updated to not rely on Bootstrap or Unify. It has also been brought in line with the new typography styles and breakpoints that have been set out.


src/assets/custom/css/_sass/_service-line-case-studies.scss
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+  @include medium-and-large {
+    margin-bottom: 40px;
+    max-width: 300px;
+  }
+  @include medium-large-and-extra-large {
+    margin-left: $lateral-space--small;
+    margin-right: $lateral-space--small;
+  }
```
LL782: Looks like the margins can be on the generic. Maybe the max-width as well?

LL782 (approved):
Code looks all good! Haven't checked it in a browser myself but assuming it's good in the eyes of the designer let's get this in

LL782:
Had a look in the browser myself. Looks all good other than at the wider end of small (e.g. 767px wide) where I wonder if a max width on the cards would be good. Perhaps run it by @anguspaterson

src/assets/custom/css/_sass/_service-line-case-studies.scss
Dan-Bird: That's a good spot with the Margins. I'll run the max width on small by Angus

LL782:
https://github.com/codurance/site/pull/1844/commits/0788899cf7be8c56d9c8ea39682a2838f0cb6211 looks good :+1:

PR #1845, opened 2020-06-22
Add Testimonials section to Service Line Page Template

Dan-Bird:
This PR adds the Testimonials section to the above template in order to be applied across all the service lines.


LL782 (approved):
:+1: looks good!

PR #1846, opened 2020-06-22
Removed Karolina profile from Codurance website

ghost:
- [ X ] Remove Karolina profile as she no longer works for Codurance since last 1st May 2020


LL782 (approved):
:+1:

PR #1848, opened 2020-06-23
Update links back to Softmod

Dan-Bird:
This PR adds links back to the Software Modernisation page within Soft-Mod articles.


LL782 (approved):
🏈

PR #1850, opened 2020-06-24
Updates to Service Line Tabber

LL782:
This work is part of including the Tabber in the Service Line Details layout. See [this LeanKit card](https://codurance-online.leankit.com/card/1110231897) for more details.

Other clean-up that happened:

- As we replaced the old tabber we were able to delete a lot of CSS that is not being used anymore
- We cleaned up the `announcements_slice` previously `sm_cta_slice` as it was the only thing left using the remaining CSS
- Fixed broken background images
- Made all the SVGs the same size and removed CSS that varied from tab to tab
- Decoupled the tabber from Bootstrap and Unify and brought it inline with our BEM and [scoped styles](https://css-tricks.com/saving-the-day-with-scoped-css/) conventions
- Used our own typographic sizes and spacings
- Adapt JS to allow for the variable content
- Adapt Liquid to allow for variable lenghts of lists and tabs

These changes complete the Service Line layout which replaces the hardcoded Software Modernisation code 

Design review with @anguspaterson has been completed


Dan-Bird:
I've noticed the Intro text on the tabs has a full stop at the end in ES. In EN, however, there is only a full stop on the first one. Not something I would necessarily consider blocking - but something worth fixing for consistency.

src/assets/custom/css/main.scss
```diff
@@ -21,6 +22,7 @@ $vertical-overlap: 100px;
 @import "page-location-anchor";
 @import "section-intro";
 @import "service-line-teaser-group";
+@import "tabber";
```
mattgraygithub: Should this be in alphabetical order?

mattgraygithub (approved):
Looks good to me 🎉 🌮

src/assets/custom/css/_sass/_announcement-slice.scss
```diff
@@ -0,0 +1,52 @@
...
+.announcement-slice__icon {
+  fill: transparent;
+  float: left;
+  margin-right: 15px;
+  stroke-width: 5px;
+  stroke: white;
+  width: 30px;
+}
+
+.announcement-slice__link {
```
Dan-Bird: If this is not used then we can delete it 🔪

Dan-Bird (approved):
All good. 

Looks a lot better 👍

src/assets/custom/css/_sass/_announcement-slice.scss
LL782: Cool. Resolved here: https://github.com/codurance/site/pull/1850/commits/383cafc908213c54ec17b2ab0f937632b8bf2f2f

src/assets/custom/css/main.scss
LL782: Good call. Fixed here: https://github.com/codurance/site/pull/1850/commits/49fe0d0bab134182875522e16c01bf5db783c059

PR #1851, opened 2020-06-24
Ensure tab has a 50/50 split of content width

LL782:
This fixes an issue with the blue and white areas not always being the same width.

See **before** and **after** shots here: 
Screenshot 2020-06-24 at 15 49 21

Screenshot 2020-06-24 at 15 48 56


mattgraygithub (approved):
👍 👍 👍

PR #1852, opened 2020-06-24
Tabber copy refinement

LL782:
Here we have:
- Applying title-casing to text in the tabber
- Replacing hyphens with "non-breaking-hyphens"
- Correcting some spaces before and after semi-colons

### DIFF
--- a/src/_data/service_lines_tabber.yml
+++ b/src/_data/service_lines_tabber.yml
@@ -102,7 +102,7 @@ es:
 
   software-modernisation:
     - id: cambio-sostenible
-      title: Cambio sostenible
+      title: Cambio Sostenible
       intro: Las empresas podrán evolucionar rápidamente si les ofrecemos software de calidad.
       list:
         - Reducir el tiempo de entrega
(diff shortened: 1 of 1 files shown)


src/_data/service_lines_tabber.yml
```diff
@@ -133,34 +133,34 @@ es:
         - Reducción de costes
         - Actualización tecnológica
         - Mitigación frente al end-of-life
-      details: Cloud native; arquitectura sin servidor; integración de terceros; implementación en contenedores; escalamiento elástico; seguridad; automatización;Infrastructure-as-code; Managed databases; Modern frameworks Modular architecture
+      details: Cloud native; Arquitectura sin servidor; Integración de terceros; Implementación en contenedores; Escalamiento elástico; Seguridad; Automatización; Infrastructure&#8209;as&#8209;code; Managed databases; Modern frameworks; Modular architecture
       image-path: /assets/custom/img/service-lines/software-modernisation/tabber-technology.svg
 
     - id: aproximacion-de-negocio
-      title: Aproximación de negocio
+      title: Aproximación de Negocio
```
Dan-Bird, line 140: I would check with Helena on capitalising more than the first word in Spanish. I think they have different capitalisation rules.

LL782: I hear you. There's this thread in Slack. https://codurance.slack.com/archives/C0136BDL3FW/p1593010949028000
We can always change it back if needed but for now @nicholemellekas has asked for it and we've found some rules to suggest it will be ok

Dan-Bird: 👍

Dan-Bird (approved):
Seems all good to me

PR #1854, opened 2020-06-25
Add translated copy for services page clients teaser

mattgraygithub:
Fixes the issue of not having the clients teaser title translated to Spanish on the services overview page


Dan-Bird (approved):
This seems good to me

LL782 (approved):
:+1:

PR #1855, opened 2020-06-25
Add software delivery page

mattgraygithub:
This adds an index page for the software delivery service line, as per [leankit ticket 221](https://codurance-online.leankit.com/card/1069431621) and adds all of the copy.

Also did some cleanup of the translations by removing duplicate entries for the 'How we solve your problems' section and the contact CTA's, since the copy for these is the same across all service lines.

Also added the Software Modernisation publications 

Notes/still to do for the software delivery page overall:
- Change tabber nav bar styling so that the nav items stretch across the entire block
- Find out about design changes to text in hero section and make changes if necessary 
- Add icons and images
- Add link to this page from services page
- change metarobots and sitemap info


Dan-Bird (approved):
Changes look good to me. 

From looking at the page in the browser there are a few things to tidy up on the template and tabber after this is merged in.

src/services/software-delivery/index.html
```diff
@@ -0,0 +1,22 @@
...
+metarobots: noindex,nofollow
+sitemap: false
+
+intro:
+  background-image-small: /assets/custom/img/service-lines/service-line-example/meta-social-image.png
+  background-image-large: /assets/custom/img/service-lines/service-line-example/meta-social-image.png
+  foreground-image-large: /assets/custom/img/service-lines/service-line-example/meta-social-image.png
+
+curated_publications:
+location: 'soft-mod'
```
LL782: What do you think about changing this location to `software-delivery-page` or something similar?

We would then have to add that location to the posts that we want to show in this location but I think it keeps it more understandable if each page is it's own location.

LL782 (approved):
Code looks good. I haven't checked in the browser. 

One comment but doesn't need to be addressed before merging necessarily.

src/assets/custom/css/_sass/_services-curated-publications.scss
```diff
@@ -6,7 +6,7 @@
   }
 }
 
-.services__curated-publications--one-publication {
+.services__curated-publications--small-margin-top {
```
LL782: Not sure about this rename? Want to make a case for it? Not a biggie if you want to merge but I think the classname should describe the what the element is not how it should be styled.

LL782 (approved):
Just re-reviewed and all the latest commits look good! LGTM

src/assets/custom/css/_sass/_services-curated-publications.scss
mattgraygithub: Struggling to think of another name! - it's basically the curated publications if there's only one publication OR if there is no webinar. Any ideas?

PR #1864, opened 2020-06-29
Service line template/update intro section

Dan-Bird:
This PR updates the styling of the Intro Section within the Service Line Template to bring it inline with the latest designs


mattgraygithub (approved):
Looks good 👍

PR #1865, opened 2020-06-29
Slt 464/nav hide and reveal

cherryt:
We have decided to use `position: sticky` for the website header and have turned the feature toggle on, because to have the current header functionality without JS. We have tested this in IE, which falls back to a static navigation bar.


AurelijaZuba (approved):

Dan-Bird (approved):
Like the `sticky` positioning

👍

PR #1869, opened 2020-06-30
Make publications show on small screen 

mattgraygithub:
This fixes is the issue of publications in service line pages being hidden on a small screen if there is only one publication or there is no webinar on the page


LL782 (approved):
Looks sane to me!

PR #1879, opened 2020-07-03
SLT-464 Smooth header animation

LL782:
This is further work towards **[SLT-464 Reveal and hide navigation based on scrolling up and down](https://codurance-online.leankit.com/card/1136109509)**

Done:

- introduce a tolerance to the scroll listener, so the header only responds to a _decisive scroll_ and we don't run the risk of it appearing and disappearing in quick succession. This tolerance has also been added to the "atTop" concept to ensure the header always returns to it's natural position when it reaches the top of the page

- add throttling on the scroll listener, so the rate at which we look to make updates is controlled by the availability of animation frames according to the browser

- when the header has been revealed by scrolling up it is now hidden by scrolling down 

- lots of clean up in the test file for improved readability and maintenance 

To be completed in a later branch:

- When the header hides or reveal, ensure it _slides_ out of view over the course of 400ms. Currently it will just appear and disappear in an instant.

This work is behind feature toggle `slt-464-new-js` (off) but could potentially be released at this stage.


src/assets/custom/js/websiteHeader.test.js
```diff
@@ -37,27 +29,84 @@ describe("Website Header", () => {
...
+}
+
+function simulateScrollingToY(y) {
+  window.scrollY = y;
+
+  window.document.dispatchEvent(
+    new Event("scroll", { bubbles: true, cancelable: true })
+  );
+
+  jest.runAllTimers();
```
Dan-Bird: We could store this in a small function describing the use, i.e triggerMockRequestAnimationFrame

Dan-Bird (approved):
Left a comment, but still all good for me 👍

src/assets/custom/js/websiteHeader.test.js
LL782: Nice one, I've done this https://github.com/codurance/site/pull/1879/commits/a7990b27df8d3eefd383867bf9eaaa33d84af18c

PR #1881, opened 2020-07-03
Ensure there is space for the +/- icon

LL782:
Ensure the tabber title always leaves space for the +/- icon on small screens.

This fixes bugs like the one described here:
https://codurance-online.leankit.com/card/1150256445

**Before**
Screenshot 2020-07-03 at 14 41 24

**After**
Screenshot 2020-07-03 at 14 56 51


Dan-Bird (approved):
This works for me

PR #1889, opened 2020-07-06
Amend icon on announcement slice and update image on SoftMod CTA banner

anguspaterson:
Amend icon on the slice to reflect new design and update image on in-page banner. Max width has also been added to in-page banner to 1300px (new max-width)


src/assets/custom/css/_sass/_in-page-banner.scss
```diff
@@ -3,6 +3,8 @@ $banner-height: 300px;
 .in-page-banner {
   min-height: $banner-height;
   overflow: hidden;
+  max-width: 1300px;
```
Dan-Bird, line 6:
```suggestion
  max-width: $extraLargeMin;
```

Dan-Bird: Could possibly use this variable - but I guess this will need adjusting again when the Homepage get's fully reworked

Dan-Bird (approved):
Seems good

src/assets/custom/css/_sass/_in-page-banner.scss
anguspaterson: sounds good - let's save this for a future iteration

PR #1891, opened 2020-07-06
Change orange to colour on the majority of remaining buttons and links

anguspaterson:
There will be a future commit for backgrounds to remove green and reset orange to coral. This commit does not cover Katylist.


Dan-Bird (approved):
Makes sense

LL782 (approved):

PR #1899, opened 2020-07-13
Cazoo links

LL782:
Update links to Cazoo to use the www subdomain, as requested via @haletothewood. This affects the clients page and the link from the Serverless podcast:

https://codurance.com/clients
https://codurance.com/podcasts/2019-12-02-serverless/

In the Serverless podcast I've added a comma between the two links to make them clearer to read.

Lastly, there was a closing [`sub`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/sub) tag in the HTML for the podcast with no opening tag. I have corrected this, using a [`small`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/small) tag which seems more appropriate for accreditation of the photo and music.

### DIFF
--- a/src/_data/clients.yml
+++ b/src/_data/clients.yml
@@ -34,7 +34,7 @@
   name: Cazoo
   image: /assets/custom/img/home/clients/cazoo-logo.png
   alt-text: Cazoo Logo
-  website: cazoo.co.uk
+  website: www.cazoo.co.uk
   services:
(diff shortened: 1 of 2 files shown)


haletothewood (approved):
Thanks Laurence. 👍 . The `www` should be enough but I'll circle back if they require `https`.

PR #1922, opened 2020-07-21
SLT 555 - Hero section training page example

mattgraygithub:
Work in progress for [training page example hero section](https://codurance-online.leankit.com/card/1172891846)

Changes will be done as part of other tickets: 
- CTA href
- Remove hard coded data
- Sandro's icon and background image to be changed

Better naming suggestions will be gladly accepted. And if any other refactoring bits have been missed.


src/assets/custom/css/_sass/_training-hero.scss
```diff
@@ -0,0 +1,150 @@
...
+  @include medium-and-large {
+    height: 433px;
+  }
+
+  @include extra-large {
+    height: 519.5px;
+  }
+}
+
+.trainint-hero__course-instructor-avatar {
```
cherryt: Is this a typo? :D (Also in the liquid file this class is used)

AurelijaZuba: good spot! thanks it is a typo 😱

src/assets/custom/css/_sass/_training-hero.scss
```diff
@@ -0,0 +1,150 @@
...
+  @include medium-and-large {
+    height: 433px;
+  }
+
+  @include extra-large {
+    height: 519.5px;
+  }
+}
+
+.training-hero__course-instructor-avatar {
```
cherryt: Should this be moved down to be with the other -course-instructor- classes? :)

AurelijaZuba: Probably yes :D moved now 🎊

src/_includes/training_hero.liquid
```diff
@@ -0,0 +1,24 @@
...
+      </div>
+      <div class="training-hero__intro-container">
+        <p class="training-hero__intro-description">This course is designed to help developers write code that is well-crafted. Meaning code that is clean,
+        testable, maintainable, and an expression of the business domain. The course is entirely hands-on,
+        designed to teach developers practical techniques they can immediately apply to real-world projects.
+        </p>
+        {% include cta_primary_link.html color='coral' text="Enquire now" %}
+      </div>
+    </div>
+</section>
```
cherryt, line 33: In this file, it looks like there are 3 sections. Would each of those sections have its own block name? 
For example, training-hero__course-instructor-* is repeated a few times, could the block be 'training-hero-course-instructor'? ie, training-hero-course-instructor__avatar
Just an observation, not totally sure myself :)

cherryt (approved):
Good work :)

src/assets/custom/css/_sass/_training-hero.scss
```diff
@@ -0,0 +1,149 @@
+.training-hero__background-image {
+  background-size: cover;
+  opacity: 0.9;
+  position: relative;
+  width: 100%;
```
Dan-Bird: We should include our max-width includes here.

src/assets/custom/css/_sass/_training-hero.scss
```diff
@@ -0,0 +1,149 @@
+.training-hero__background-image {
+  background-size: cover;
+  opacity: 0.9;
+  position: relative;
```
Dan-Bird: Is this needed?

mattgraygithub: True. Not anymore I don't think

src/_includes/training_hero.liquid
```diff
@@ -0,0 +1,24 @@
+<section class="training-hero">
+    <div class="training-hero__background-image" style="background-image: url({{ site.baseurl }}{{ page.intro.background-image }})"> </div>
+    <div class="training-hero__lower-background-block"></div>
+    <div class="training-hero__content-container">
+      <div class="training-hero__course-content-info-container">
+        <h1 class="training-hero__title">Crafting Code</h1>
+        <span class="training-hero__duration">Duration: 2 days</span>
+      </div>
+      <div class="training-hero__course-instructor-container">
+        <span class="training-hero__course-instructor-avatar" style="background-image: url({{ site.baseurl }}{{ page.intro.instructor-avatar }})"></span>
```
Dan-Bird: This background doesn't appear.

src/assets/custom/css/_sass/_training-hero.scss
mattgraygithub: Yep I agree

src/_includes/training_hero.liquid
```diff
@@ -0,0 +1,24 @@
+<section class="training-hero">
+    <div class="training-hero__background-image" style="background-image: url({{ site.baseurl }}{{ page.intro.background-image }})"> </div>
+    <div class="training-hero__lower-background-block"></div>
```
Dan-Bird: I'm not sure this will need a separate div just for the background. I think we can re-organise the structure of this to be a bit cleaner and possibly use fewer elements.

src/assets/custom/css/_sass/_training-hero.scss
```diff
@@ -0,0 +1,149 @@
...
+  width: 100%;
+
+  @include small {
+    height: 216.27px;
+  }
+
+  @include medium-and-large {
+    height: 433px;
```
Dan-Bird: We've been rounding measurements to the nearest 5px.

src/assets/custom/css/_sass/_training-hero.scss
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  @include extra-large {
+    height: 519.5px;
+  }
+}
+
+$background-block-height: ($vertical-overlap *  4);
```
Dan-Bird: I wouldn't use the `$vertical-overlap` var here if its not concerning itself with any vertical overlap.

src/assets/custom/css/_sass/_training-hero.scss
```diff
@@ -0,0 +1,149 @@
...
+.training-hero__lower-background-block {
+  background-color: $slate;
+  position: relative;
```
Dan-Bird: I'm not sure if this is needed?

src/assets/custom/css/_sass/_training-hero.scss
```diff
@@ -0,0 +1,149 @@
...
+
+  @include small {
+    height: 2px + $background-block-height;
+  }
+
+  @include medium {
+    height: 6px + $background-block-height;
```
Dan-Bird: I'm not clear why we need to amend the height in this manner? I think we could simplify this when we re-arrange the HTML markup.

src/assets/custom/css/_sass/_training-hero.scss
```diff
@@ -0,0 +1,149 @@
...
+  position: relative;
+  color: white;
+  transform: translateY(-650px);
+
+  @include small {
+    padding-top: 85px;
+  }
+  
+  @include small-and-medium {
+    @include max-width--out-there;
```
Dan-Bird: This would set a max width larger than our medium breakpoint. We can take a look at this.

src/assets/custom/css/_sass/_training-hero.scss
```diff
@@ -0,0 +1,149 @@
...
+.training-hero__course-instructor-container{
+  display: flex;
+  padding: 50px 0;
+}
+
+.training-hero__course-instructor-avatar {
+  background-position: center;
+  background-size: cover;
+  height: $vertical-overlap + 10px;
+  width: $vertical-overlap;
```
Dan-Bird: As previous, I wouldn't use this var unless setting a vertical overlap.

src/assets/custom/css/_sass/_training-hero.scss
```diff
@@ -0,0 +1,149 @@
...
+  @include mercer-bold;
+  color: white;
+}
+
+.training-hero__intro-container {
+  @include small {
+    @include base;
+
+    & .cta--primary-coral {
+      margin-right: 27px;
```
Dan-Bird: We shouldn't need this. The gap should come from padding within it's container. The reason you need this margin on the right is because `training-hero__content-container` has a margin left pushing the content away from the left.

src/assets/custom/css/_sass/_training-hero.scss
```diff
@@ -0,0 +1,149 @@
...
+    @include freed;
+  }
+
+  @include large-and-extra-large {
+    & .cta--primary-coral {
+      display:none;
+    } 
+  } 
+
+  & .training-hero__intro-description {
```
Dan-Bird: We don't need to nest this. It has it's own class to target specifically and it's style doesn't relate to anything within the parent selector.

src/_includes/training_hero.liquid
AurelijaZuba: Sigh... not sure why it works locally. Pushing some changes so hopefully that appears

Dan-Bird (commented):
I think there are a few things we could look at tidying up - happy to go through them.

Dan-Bird (approved):
Lets get this in and start working on including the form 👍

PR #1925, opened 2020-07-22
Fix links overlapping announcement slice when not visible

Dan-Bird:
This PR fixes a live issue where the links within the dropdown menus are still clickable when they are not visible.


anguspaterson:
Fix is good! Great work

PR #1928, opened 2020-07-23
Add support for auto rebuild in windows docker

paulmorrishill:
In windows the containers you can bind volumes but they won't trigger file watchers, jekyll supports using a polling file watcher instead but the option has to be passed to enable it as it is more CPU intensive than normal file watching. 

I have refactored the rake file a bit to allow for this.


ghost (approved):
LGTM

PR #1934, opened 2020-07-24
Updated images to reflect correct order as per service lines

anguspaterson:
The icons we in the incorrect order and did not match the order in which the service lines were preseneted


Dan-Bird (approved):
All Good

PR #1936, opened 2020-07-27
Slt 594/service line overview ctas

cherryt:
- Add CTAs to service lines buttons, training teaser and compass teaser buttons
- Fixed example tabber in the ui-components page (https://codurance.com/ui-components/tabber/)

### DIFF
--- a/src/_data/service_lines.yml
+++ b/src/_data/service_lines.yml
@@ -14,7 +14,6 @@ software-modernisation:
   es-description: Consigue alinear las necesidades de negocio con las soluciones técnicas para permitir que los cambios se realicen de manera más sencilla y segura.
   es-hubspot-id: f913b7b4-3a24-42d0-bae8-2b76f4ff0f88
   image-src: /assets/custom/img/services/serviceline-softmod.jpg
-  cta-path: /services/software-modernisation/
 
 software-delivery:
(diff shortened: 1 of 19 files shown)


src/_data/service_lines.yml
```diff
@@ -9,51 +9,56 @@ software-modernisation:
   name: Software Modernisation
   en-title: Improve business agility, and make innovation possible
   en-description: Align your business and technical vision to enable change to be made more simply and securely.
+  en-hubspot-id: 57099049-6f33-4639-aa6d-e86e437a37fd
```
Dan-Bird: What do you think on naming this `en-hubspot-cta-id` ? 

My thoughts around the change are just that we currently use Hubspot IDs for forms and CTAs now. Maybe its good to start differentiating them.

src/_i18n/en.yml
```diff
@@ -197,8 +197,7 @@ services:
   training:
     title: Training
     description: Build a team you can trust to perform. Help your team build the skills they need to excel in every area of your product delivery. Our experts share years of experience building mission-critical systems in a wide range of technologies.
-    hubspot-form-id: 43dee6bc-24ce-4289-82cf-36f014d40be4
-
+    hubspot-id: becaa5e5-8a5d-4358-8b49-b8d8bb572308
```
Dan-Bird: Same point as above here and throughout.

Dan-Bird (approved):
Small note on the naming of some stuff, but otherwise it looks good. 

Nice work!

src/_data/service_lines.yml
cherryt: Ya the -cta-id ending would be nicer, I just kept with convention thats used in other areas, but I'll rename all of them to make it more clear :)

PR #1940, opened 2020-07-28
SLT-577 Close Sub-menus on Scroll

Dan-Bird:
This PR improves usability of the Website Header by adding functionality that closes sub-menus when you scroll down. 

There are a few changes such as using `position: fixed` for the Header instead of `position: sticky` which is more compatible with IE.

This also mitigates two issues we were facing with the Header being:

1) Janky opening and closing animation with sub-menus
2) A bug noted [here](https://codurance-online.leankit.com/card/1177537514) - where scrolling down a couple of pixels and opening a sub-menu would cause the header to move upwards off-screen.


src/assets/custom/css/_sass/_website-header.scss
```diff
@@ -11,15 +28,27 @@ body:not(.toggle--slt-397-website-header-updates) .website-header {
...
 .toggle--slt-397-website-header-updates {
   .website-header {
     left: 0;
-    position: sticky;
+    position: fixed;
+    top: 0;
     width: 100%;
     z-index: 100;
+    transition: transform 0.3s ease;
+    transform: translateY(-100%);
```
LL782, line 39:
```suggestion

    &:not(.website-header--revealed) {
      transform: translateY(-100%);
    }
```

src/assets/custom/css/_sass/_website-navigation.scss
```diff
@@ -75,6 +75,7 @@
     }
     &.website-navigation__menu--showing-sub-menu {
       left: -100vw;
+      overflow-y: hidden;
```
LL782, line 71: Just be careful that on a small or medium screen that doesn't have sufficient height to show the whole menu you can still scroll it down to reach the views that don't fit on screen

src/assets/custom/js/custom.js
```diff
@@ -74,43 +74,20 @@ $(window).resize(function () {
...
-(function navbarProjectsActive(){//TODO delete once we add projects to the navbar
-	$(function() {
-		const url = window.location.href;
-		$(".nav a").each(function() {
-			if (url.includes("/projects") && this.href.includes("/client")) {
-				$(this).closest("li").addClass("active");
-			}
-		});
-	});
-})();
```
LL782, line 98: Are these needed if the feature-toggle is `off`?

Incidentally when is the feature-toggle going to be removed?

src/assets/custom/js/websiteHeader.js
```diff
@@ -1,19 +1,19 @@
 var websiteHeader = function () {
-  var HEADER_SELECTOR = ".website-header";
-  var REVEALED_CLASS = "website-header--revealed";
-  var HEADER_HAS_OPEN_SUBMENU_CLASS = "website-header--has-open-submenu";
-  var tolerance = 2;
+  window.__CODURANCE = window.__CODURANCE || {};
+  window.__CODURANCE.websiteNavigation = __CODURANCE.websiteNavigation || {};
 
-  var previousWindowPosition;
+  var closeSubMenu;
```
LL782: Why is this here? Why not move it closer to where it's used

src/assets/custom/js/websiteHeader.js
```diff
@@ -31,42 +31,59 @@ var websiteHeader = function () {
...
-    if (startingPositionIsInView(latestWindowPosition)) {
-      return;
-    }
+    if (downwardMovement && !websiteHeaderOpen()) {
+      if (subMenuOpen) {
+        closeSubMenu =
+          window.__CODURANCE.websiteNavigation.closeOpenSubMenu ||
+          function () {};
+        closeSubMenu();
+      }
```
LL782:
```suggestion
      if(
        window.__CODURANCE && 
        window.__CODURANCE.websiteNavigation && 
        window.__CODURANCE.websiteNavigation.closeOpenSubMenu
      ) {
        window.__CODURANCE.websiteNavigation.closeOpenSubMenu();
       }
```

Personally I wouldn't assume that `window.__CODURANCE.websiteNavigation.closeOpenSubMenu` is defined. 

Also, that function already handles whether or not a subMenu is open so you can remove `subMenuOpen` from here and potentially from this whole file. Which would be nice.

src/assets/custom/js/websiteHeader.js
```diff
@@ -31,42 +31,59 @@ var websiteHeader = function () {
...
       return;
     }
 
-    if (startingPositionIsInView(latestWindowPosition)) {
-      return;
-    }
+    if (downwardMovement && !websiteHeaderOpen()) {
+      if (subMenuOpen) {
+        closeSubMenu =
+          window.__CODURANCE.websiteNavigation.closeOpenSubMenu ||
```
cherryt: Just curious why an empty function call is needed, would it possible to not call a function at all if websiteNavigation.closeOpenSubMenu is undefined? :)

src/assets/custom/js/websiteHeader.test.js
```diff
@@ -1,120 +1,139 @@
...
+          expect(header.classList).toContain('website-header--revealed');
           expect(header.classList.length).toBe(2);
         });
 
-        describe("When we then scroll close to the top of the page", () => {
-          beforeEach(() => {
-            simulateScrollingToY(1);
+        describe('When we then open a submenu', () => {
+          beforeAll(() => {
+            header.classList.add(HEADER_HAS_OPEN_SUBMENU_CLASS);
```
LL782: If you move the responsibility to know if a submenu is open out of this file, as per the previous comment/suggestion then you don't need to test both branches (submenu open + submenu not open). That would be simplier

src/assets/custom/js/websiteHeader.test.js
```diff
@@ -1,120 +1,139 @@
...
-          beforeAll(() => {
-            header.classList.add(HEADER_HAS_OPEN_SUBMENU_CLASS);
-          });
-
-          describe("And we scroll down the page by three pixels", () => {
+        describe('when a submenu is not open', () => {
+          describe('And we scroll down the page by three pixels', () => {
             beforeEach(() => {
-              simulateScrollingToY(1000);
+              simulateScrollingToY(1000 + 100);
```
LL782: The description talks about "three pixels" because we were testing the boundaries. Now this arrangement is introducing much bigger steps. Careful to keep it readable and understandable

src/assets/custom/css/_sass/_website-header.scss
Dan-Bird: This reads much better 👍

src/assets/custom/js/websiteHeader.test.js
```diff
@@ -1,120 +1,139 @@
...
+        it("we let the header return to it's natural position", () => {
+          expect(header.classList).toContain('website-header');
+          expect(header.classList).toContain('website-header--revealed');
+          expect(header.classList.length).toBe(2);
+        });
+
+        describe('When we then scroll down the page by a couple of pixels', () => {
           beforeEach(() => {
-            header.classList.remove(HEADER_HAS_OPEN_SUBMENU_CLASS);
+            simulateScrollingToY(3);
```
LL782: This isn't really part of these changes, but the more times we look at these tests the more I recognise the benefit of having a `simulateScrollingBy` function to go with `simulateScrollingToY`

Something like:
```javascript
function simulateScrollingBy(diff) {
  window.scrollY = window.scrollY + diff;

  window.document.dispatchEvent(
    new Event('scroll', { bubbles: true, cancelable: true })
  );


  triggerMockRequestAnimationFrame();
}
```

LL782 (approved):
Left a bunch of comments. Generally all looks really good. Only major thing is I wouldn't assume the existence of the __CODURANCE object in global scope. Without integrated tests of some sort to protect you I would code those bits a little more defensively, e.g.

```javascript
if(
  window.__CODURANCE && 
  window.__CODURANCE.websiteNavigation && 
  window.__CODURANCE.websiteNavigation.closeOpenSubMenu
) {
  window.__CODURANCE.websiteNavigation.closeOpenSubMenu();
}
```

src/assets/custom/css/_sass/_website-navigation.scss
Dan-Bird: Yup, the menus themselves still scroll if they get cut off 😄

src/assets/custom/js/websiteHeader.js
Dan-Bird: This var has been removed now since it's not needed. https://github.com/codurance/site/pull/1940/commits/bbea935a1c2a48d0f6f53ee09248e47647187d59

src/assets/custom/js/websiteHeader.js
Dan-Bird: This is a lot better and cleaner too. It's now been implemented here: https://github.com/codurance/site/pull/1940/commits/bbea935a1c2a48d0f6f53ee09248e47647187d59

src/assets/custom/js/websiteHeader.js
Dan-Bird: This is now taken care of in this commit: https://github.com/codurance/site/pull/1940/commits/bbea935a1c2a48d0f6f53ee09248e47647187d59

The empty function is to stop the script erroring and breaking if `window.__CODURANCE.websiteNavigation.closeOpenSubMenu` wasn't defined.

src/assets/custom/js/websiteHeader.test.js
Dan-Bird: This has been addressed here: https://github.com/codurance/site/pull/1940/commits/d5ac023d4dc7a8f811b5b063a477250bf415ed33

src/assets/custom/js/websiteHeader.test.js
Dan-Bird: That's a good idea. I think it allows the tests to be read a bit better. It's been implemented here: https://github.com/codurance/site/pull/1940/commits/82ac868abefbcd8d773ec6fea2cd5d184a094e8c

src/assets/custom/js/custom.js
Dan-Bird: Good point. 

Since this ticket completes the epic, I will also remove the feature toggle's as part of this PR.

mattgraygithub (approved):
Looks sensible 👍

PR #1944, opened 2020-07-31
Slt 558 add accordion single training

mattgraygithub:
This adds an [accordion section](https://codurance-online.leankit.com/card/1172910680) to the [single training example page](https://codurance-online.leankit.com/card/1151346123)/crafting code and crafted design course page.

Note - the data used is for the Crafting Code & Crafted Design course page (as we are focussing on that one first so that we can add the CTA for it into the training overview page). 

Still to do:
- Make a Crafted Code & Crafted Design index page and copy the code from this example page to that page
- Make this accordion and it's inner components work for any training page - I suggest having the course title in the front matter of each page and then changing the first line of each of the 'training_content....' includes files so that the `crafting-code-and-crafted-design` part comes from the front matter. eg. `{% assign data = include.accordion_data.[page.course-title]. why-attend %}`, then....
- Add the data for all the other courses into the training_accordion yml files
- Fix auto scrolling to account for header automatically disappearing


Dan-Bird (approved):
Seems alright. At some point, we can look at if there are cleaner ways to insert the markup and data into the accordion.

src/assets/custom/js/accordion.js
```diff
@@ -0,0 +1,74 @@
+(function() {
+
+  var ACTIVE_CLASS = "active";
+
+  var SELECTORS = {
+    ACCORDION: "[data-accordion]",
+    PANEL: "[data-accordion_panel]",
+    PANEL_ACTIVE: "[data-accordion_panel]." + ACTIVE_CLASS
```
cherryt, line 8: Is it possible to use template literals/strings here? :)

src/assets/custom/js/accordion.js
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  var ACCORDION = window.document.querySelector(SELECTORS.ACCORDION);
+
+  if (ACCORDION === null) {
+    return;
+  }
+
+  var PANELS = nodeListToArray(ACCORDION.querySelectorAll(SELECTORS.PANEL));
```
cherryt, line 17: Could this be moved closer to where its used?

src/assets/custom/js/accordion.js
mattgraygithub: Maybe! Maybe @Dan-Bird can have a look at this

cherryt (approved):

src/assets/custom/js/accordion.js
cherryt: Ya, no worries, I've approved it anyways :)

PR #1945, opened 2020-07-31
Add original colours to SVG swoosh elements

anguspaterson:
This change provides a temporary design revision ahead of future brand alignment


Dan-Bird (approved):
All good with me

PR #1949, opened 2020-08-05
SLT 558 Training Accordion Fix

Dan-Bird:
This PR fixes an issue when the accordion panels would collapse if the content of the panel is clicked.

Now panels will only open or collapse if the panel header is clicked.


AurelijaZuba (commented):
Looks great!  🚀

AurelijaZuba (approved):

cherryt (approved):
Nice :D

PR #1952, opened 2020-08-10
Tweaks to training accordion

mattgraygithub:
This adds some functionality tweaks to the accordion in the individual training page template to add the animated plus and minus icons and adjust the autoscrolling of tabs when opened to account for the automatic hiding and showing of the website header


Dan-Bird (approved):
Good with me 👍

PR #1955, opened 2020-08-12
Slt 531/single training bio section

mattgraygithub:
This adds a [course instructor biography](https://codurance-online.leankit.com/card/1151315532) section to the [single training example/template page](https://codurance-online.leankit.com/card/1151346123).

We still need to pass the content in from yml files rather than hard coding it; and add one final design icon, but these can be done on a separate branch because we need to do this for all elements of the page and this is just an example page and not linked from anywhere yet.


src/assets/custom/css/_sass/_training-bio.scss
```diff
@@ -0,0 +1,149 @@
...
+
+.training-bio__instructor-title,
+.training-bio__instructor-name {
+  color: black;
+  display: block;
+  margin-left: 6px;
+}
+
+.training-bio__instructor-title {
+  @extend .training-hero__instructor-title;
```
Dan-Bird: We might want to avoid extending styles from a different stylesheet.

Dan-Bird: Amended here https://github.com/codurance/site/pull/1955/commits/ec66d1198e3a8dc93e1fc46a30c81cd154f2679f

PR #1961, opened 2020-08-19
Adapt single training template to work for all training pages

mattgraygithub:
This makes our single training page template work for all courses, whether or not they include instructor info in the hero or an instructor bio section.


Dan-Bird (approved):
All Good 👍

PR #1966, opened 2020-08-21
Fix bug for card carousel

Dan-Bird:
This PR fixes a bug introduced from refactoring the trainings data files, where more navigational dots than necessary were getting created for the training card carousel.


mattgraygithub (approved):
Looks good to me 👍

PR #1967, opened 2020-08-24
Remove styles which make cta text coral or orange after being clicked

mattgraygithub:
This fixes [the bug](https://codurance-online.leankit.com/card/1202978808) of text inside the cta's on the clients page 'disappearing' (being changed to the same colour as the CTA button background) after being clicked.


Dan-Bird (approved):
This all looks good to me

# 🏎️

PR #1968, opened 2020-08-25
Fix Footer Links on Hover

Dan-Bird:
This PR fixes a bug where links would change to incorrect colours due to overrdining styles within our bootstrap and unify theme stylesheets.


mattgraygithub (approved):
BEM tidy 👍

PR #1976, opened 2020-08-27
Reduce Coupling from Unify and Bootstrap

Dan-Bird:
This PR begins to uncouple the website from using the Unify and Bootstrap libraries. 

Some dead code has been tidied up. A section of the Homepage has been cleaned to follow the BEM naming convention and reduce coupling from helper classes from the above mentioned libraries.


chriseyre2000:
❤️ ❤️ ❤️ ❤️ ❤️ 

Killing Unify!

src/index.html
```diff
@@ -84,7 +84,7 @@ <h2 class="h3">{% t index-activity.title %}</h2>
 
 
 <!-- Partner panel -->
-<div class="section gray-section">
+<section class="section gray-section">
```
mattgraygithub, line 87: Guessing we're going to do more with these classes to make them more descriptive, and then remove the comments? Looks great so far though!

mattgraygithub (approved):
Looking good to me 👍

src/index.html
Dan-Bird: Yeah exactly. Once they follow the BEM convention they'll already be descriptive and the comments would then be redundant

PR #1977, opened 2020-08-27
Setting Ignacio Saporiti's role to 'Software Craftsperson-in-training'.

isaporiti:


ghost (commented):
LGTM. Would just double check if that is the standard naming convention.

isaporiti:
Merging after having checked existing conventions.

PR #1979, opened 2020-09-01
Open training pages in new window, with no header or footer

mattgraygithub:
This makes changes to the training pages so that they don't include the website header/navigation or footer (had to do a bit of rearrangement of the default layout to make this happen), adds the codurance logo to them in the same horizontal position as the logo in the header; and make all links to training pages (in the training overview page) open in a new window

Leankit ticket [here](https://codurance-online.leankit.com/card/1250156045)


anguspaterson (approved):
looks good - we reviewed it together

PR #1996, opened 2020-09-09
Add breaking down the monolith blogpost

mattgraygithub:
Adds the Breaking Down the Monolith blog post by Matt Belcher


src/assets/custom/css/_sass/_compass-teaser.scss
```diff
@@ -23,3 +23,14 @@
...
   margin-bottom: 25px;
 }
+
+.blog-post-compass-teaser {
+  background-color: $coral;
+  background-repeat: no-repeat;
+  background-size: cover;
+  background: url("#{$site-base-url}/assets/custom/img/softmod/bg-compass-cta.jpg");
+  margin-left: auto;
+  margin-right: auto;
```
Dan-Bird: Maybe all the common areas could be styled together:

```.compass-teaser,
.blog-post-compass-teaser { ```

And then we can have a separate entry for any changes to try and reduce the repetition. We can have a call and I can explain further if need be.

File: src/_includes/compass_teaser_for_blog_posts.liquid [hunk 2]
```diff
@@ -0,0 +1,8 @@
+<div class="blog-post-compass-teaser">
+  {%- assign anchor_id = site.translations[site.lang].compass-teaser.anchor-id -%}
+  {% include page_location_anchor.html anchor_id=anchor_id %}
```
Dan-Bird: Not sure if we would need to include a page location anchor on blog posts. 

I think it would be unlikely that we would link someone to the compass cta section of a blog post.

Dan-Bird (approved):
Overall it looks good. I've left a couple of comments that could help tidy things a little.

src/_includes/compass_teaser_for_blog_posts.liquid
mattgraygithub: Ah yeah!

src/assets/custom/css/_sass/_compass-teaser.scss
mattgraygithub: I hear ya. All sorted

PR #2000, opened 2020-09-10
Fix accordion collapsing when contents are clicked

mattgraygithub:
This fixes [a bug on the service line pages ](https://codurance-online.leankit.com/card/1200399192) whereby if you click on the content of an open accordion panel, the panel closes. We don't want this. Instead we want the panel only to close if the title bar of the panel is clicked.


Dan-Bird (approved):
Looks good.

PR #2010, opened 2020-09-18
Add redirect to head to test that 404 page will perform a redirect fo…

mattgraygithub:
This is just a test to see whether redirects from a url for a page that doesn't exist, will redirect from our 404 page. I can't test this locally or on a preview branch because our 404 page is only available from the live environment.


LL782 (approved):
:+1:

PR #2017, opened 2020-09-22
Change urls based on site language

mattgraygithub:
This is to change all of the urls within the /services/ section of the website for the **spanish site only** to be /servicios/.

as per [this leankit ticket](https://codurance-online.leankit.com/card/1304785987)

To redirect the es/services/ links to es/servicios/ I had add a new folder outside of the main src folder with all of the redirects files. This is because Jekyll will always build two versions of any html in the main src folder (one for EN and one for ES). And we cannot have en/services pages redirecting to themselves! 

Because of this slight structure change, I had to modify the deployment so that this new redirects-for-spanish-services-pages folder was deployed to the correct folder in the S3 bucket **only for the Spanish site**


src/services/training/introduction-to-functional-programming/index.html
```diff
@@ -6,7 +6,11 @@
 description-es: Conocimientos básicos sobre programación funcional y sus beneficios. Ejercicios prácticos con los que el equipo podrá iniciarse en Functional Programming.
 image:
    src: /assets/custom/img/training/introduction-to-functional-programming/training-intro-functional-programming-social-share.jpg
-metarobots: index,follow
+metarobots: index,follo
+
+wnamespace: services
```
LL782:
```suggestion
metarobots: index,follow

namespace: services
```

mattgraygithub: 🤦  thanks Laurence! 👍

chriseyre2000 (approved):
LGTM

You could possibly avoid the cd's by using the relative path in the sync, but this is clearer.

Dan-Bird (approved):
👍 looks good to me

PR #2045, opened 2020-10-22
Adds Santander and MyPulse to clients page

ewansheldon:


LL782:
Looks good and sane! :+1: Have you checked with marketing whether these logos should be added to the end of the list or if they should be closer to the top?

Also, be aware that if you want the logos to appear on Services & Solutions page in the "You're in good company" module you'll need to do something more.

http://codurance-site-pr.s3-website-eu-west-1.amazonaws.com/site-add-clients-santander-mypulse/en/services/

I don't see either of those points as blocking issues however. In my opinion it looks good to merge

LL782 (approved):

ewansheldon:
Thanks @LL782 

I will speak with marketing about their positioning on the clients page and whether they should feature in "You're in good company" - there was nothing specific requested about this on the given task

But like you say these could be updated retrospectively in a different PR.

PR #2049, opened 2020-10-30
Prepare blog service line banner

Dan-Bird:
This PR adds the Services banner so that it can be implemented within blog posts. 

To implement the banner you can add the following include where the banner is to appear within the article:

`{% include services_teaser_for_blog_posts.liquid %}`

Below is a screenshot of what the banner looks like within a post on desktop and mobile screen widths.

Screenshot 2020-10-30 at 13 24 19
Screenshot 2020-10-30 at 13 29 40


LL782:
Looks great! I think it would be good to have the options to include just the banner without the preceeding text.

LL782 (approved):
👍

PR #2061, opened 2020-12-10
Update spanish site with new posts

ewansheldon:
Adds two posts that had been added to hubspot but not jekyll
Reorders highlights page to show most recent posts/pods/videos to align with hubspot homepage


Dan-Bird (approved):
Looks good. 

LGTM 

# 🏎️

